PR #68, opened 2022-03-28
SDCORE-781: AMF should send PDU rejection message when SMF doesn't respond

vthiruveedula:


gmm/handler.go
```diff
@@ -583,6 +585,9 @@ func HandleInitialRegistration(ue *context.AmfUe, anType models.AccessType) erro
 
 	if len(ue.AllowedNssai[anType]) == 0 {
 		gmm_message.SendRegistrationReject(ue.RanUe[anType], nasMessage.Cause5GMM5GSServicesNotAllowed, "")
+		ngap_message.SendUEContextReleaseCommand(ue.RanUe[anType], context.UeContextN2NormalRelease,
+			ngapType.CausePresentNas, ngapType.CauseNasPresentNormalRelease)
+		ue.Remove()
```
thakurajayL, line 590: Can you create call flow for this problem and fix ?

gab-arrobo: @thakurajayL, it looks like this comment is outdated. Correct?

thakurajayL: I would prefer to take code as is. Valid fix to cleanup N1 tunnel.

gab-arrobo: @thakurajayL, so, should we approve this PR?

gab-arrobo:
Hi @vthiruveedula, is this PRs still valid? or should I close it?

PR #73, opened 2022-05-31
KeepAlive Timer supoort

vthiruveedula:


producer/ue_context.go
```diff
@@ -64,7 +64,7 @@ func HandleCreateUEContextRequest(request *http_wrapper.Request) *http_wrapper.R
 	}
 	var createUeContextRspData *models.CreateUeContextResponse
 	var ueContextCreateErr *models.UeContextCreateError
-	ue.EventChannel.UpdateSbiHandler(ProducerHandler)
+	ue.EventChannel.UpdateSbiHandler(UeContextHandler)
```
badhrinathpa, line 67: all these procedures haven't been tested or else they would have failed right ?

badhrinathpa (approved):
+1

PR #77, opened 2022-08-02
Catch Sigterm signal

vthiruveedula:


context/transaction.go
```diff
@@ -41,6 +41,10 @@ func (tx *EventChannel) Start() {
 		case msg := <-tx.Message:
 			switch msg.(type) {
 			case NasMsg:
+				/*if tx.NasHandler == nil {
```
badhrinathpa, line 44: We had removed this right ? any reason we have back ?

ngap/dispatcher.go
```diff
@@ -62,8 +62,12 @@ func DispatchLb(remoteAddr string, msg []byte, Amf2RanMsgChan chan *sdcoreAmfSer
 			NgapMsg: pdu,
 		}
 
+		globalRANNodeID := ranUe.Ran.RanId
```
badhrinathpa, line 65: We had removed this right ? any reason we have back ? probably needs a rebase?

ngap/dispatcher.go
```diff
@@ -62,8 +62,12 @@ func DispatchLb(remoteAddr string, msg []byte, Amf2RanMsgChan chan *sdcoreAmfSer
...
+		ranUe.Ran = ran
+		ranUe.Ran.RanId = globalRANNodeID
 		ranUe.AmfUe.EventChannel.SubmitMessage(ngapMsg)
 	} else {
+		fmt.Println("ranUe nil :  ", ran.RanID())
```
badhrinathpa, line 70: rebase

service/init.go
```diff
@@ -478,6 +487,13 @@ func (amf *AMF) BuildAndSendRegisterNFInstance() (models.NfProfile, error) {
 func (amf *AMF) UpdateNF() {
 	KeepAliveTimerMutex.Lock()
 	defer KeepAliveTimerMutex.Unlock()
+	defer func() {
+		if err := recover(); err != nil {
```
badhrinathpa, line 491: In what scenario will we get here ?

vthiruveedula: catching if there is any panic.. this is the sample code to recover from panic

context/context.go
```diff
@@ -123,7 +123,7 @@ func (context *AMFContext) AllocateAmfUeNgapID() (int64, error) {
 		return -1, err
 	}
 
-	logger.ContextLog.Infof("Allocate AmfUeNgapID : ", val)
+	logger.ContextLog.Infoln("Allocated AmfUeNgapID : ", val)
```
thakurajayL, line 126: Does not look imp log. Move it to debug.

context/context.go
```diff
@@ -41,10 +41,10 @@ func init() {
 	AMF_Self().ServedGuamiList = make([]models.Guami, 0, MaxNumOfServedGuamiList)
 	AMF_Self().PlmnSupportList = make([]factory.PlmnSupportItem, 0, MaxNumOfPLMNs)
 	AMF_Self().NfService = make(map[models.ServiceName]models.NfService)
-	AMF_Self().NetworkName.Full = "free5GC"
 	//tmsiGenerator = idgenerator.NewGenerator(1, math.MaxInt32)
-	amfStatusSubscriptionIDGenerator = idgenerator.NewGenerator(1, math.MaxInt32)
+	AMF_Self().NetworkName.Full = "free5GC"
```
thakurajayL, line 45: I see only code shuffling. Did I miss the reason for doing this ?

service/init.go
```diff
@@ -333,7 +334,7 @@ func (amf *AMF) Start() {
 	}
 
 	signalChannel := make(chan os.Signal, 1)
-	signal.Notify(signalChannel, os.Interrupt, syscall.SIGTERM)
+	signal.Notify(signalChannel, syscall.SIGHUP, syscall.SIGQUIT, syscall.SIGTERM, syscall.SIGINT, syscall.SIGSEGV, syscall.SIGCHLD)
```
thakurajayL, line 337: Which signal is received? Is it good idea to catch SIGSEGV as well ?. Do we send de-registration even during seg mentation fault ?

PR #80, opened 2022-09-02
Fixing the default mongodb path

thakurajayL:
if mongodb config was set then URL was not getting set as per configuration


badhrinathpa (approved):
+1

PR #82, opened 2022-09-22
support for multiple SMF profiles

UpadhyayaAnkur:


consumer/sm_context.go
```diff
@@ -24,16 +25,67 @@ import (
 	"github.com/omec-project/openapi/models"
 )
 
+func getServingSmfIndex(smfNum int) (servingSmfIndex int) {
+	servingSmfIndexStr := os.Getenv("SERVING_SMF_INDEX")
```
thakurajayL, line 29: Why we are using ENV variable for this ? Each AMF instance can have its own global variable which can do round robin on the list of SMFs.

consumer/sm_context.go
```diff
@@ -24,16 +25,67 @@ import (
...
+	//iterate over nf instances to ignore failed NF
+	for _, inst := range smCtxt.SmfProfiles {
+		if inst.NfInstanceId != ignoreSmfId {
+			altSmfInst = append(altSmfInst, inst)
+		}
+	}
+
+	if len(altSmfInst) > 0 {
+		smCtxt.SmfProfiles = altSmfInst
+		nfProfile := altSmfInst[0]
```
thakurajayL, line 49: - IF we always take 0th NF profile then why we create all alternate SMF list ?

consumer/sm_context.go
```diff
@@ -24,16 +25,67 @@ import (
...
+	nrfUri := ue.ServingAMF().NrfUri
+	param := Nnrf_NFDiscovery.SearchNFInstancesParamOpts{
+		ServiceNames: optional.NewInterface([]models.ServiceName{models.ServiceName_NSMF_PDUSESSION}),
+		Dnn:          optional.NewString(smCtxt.Dnn()),
+		Snssais:      optional.NewInterface(util.MarshToJsonString([]models.Snssai{smCtxt.Snssai()})),
+	}
+
+	result, err := SendSearchNFInstances(nrfUri, models.NfType_SMF, models.NfType_AMF, &param)
+	if err != nil {
+		return nil
```
thakurajayL, line 69: Log the error.

context/sm_context.go
```diff
@@ -39,6 +39,8 @@ type SmContext struct {
 	// for duplicate pdu session id handling
 	UlNASTransportVal *nasMessage.ULNASTransport
 	DuplicatedVal     bool
+
+	SmfProfiles []models.NfProfile
```
thakurajayL, line 43: Temp solution. We need to use NRF cache lib code.

consumer/sm_context.go
UpadhyayaAnkur: In the code above, we maintain only good SMF instances in amf context and remove non-reachable SMF instances(ignored one). So, it is safe to take 1st(0th index) element from remaining alternate SMF instances

consumer/sm_context.go
UpadhyayaAnkur: This code not used as of now and shall be later employed for NRF lib. So, bound to change.

context/sm_context.go
UpadhyayaAnkur: Yes, this is temporary fix until NRF cache lib codes is not available

consumer/sm_context.go
UpadhyayaAnkur: This is temporary solution, with global SMF list via NRF cache changes, this code shall be removed.

UpadhyayaAnkur:
retest this please

thakurajayL (approved):

PR #83, opened 2022-09-22
Fixing compilation issues in AMF

thakurajayL:
- Refer to correct util module
- Remove MongoDBlib & drsm replace


go.mod
```diff
@@ -2,10 +2,6 @@ module github.com/omec-project/amf
 
 go 1.18
 
-replace github.com/omec-project/MongoDBLibrary => ./MongoDBLibrary
```
badhrinathpa, line 5: are we updating mongolibrary to util path now or later?

thakurajayL: No we are still using old MongoDBLibrary. But I don't know why we had this replace path in AMF. it fails then. Looks like Vijaya missed it and it got committed.

badhrinathpa (approved):
+1

UpadhyayaAnkur (commented):
Approved

PR #91, opened 2022-11-25
Metric Function changes

UpadhyayaAnkur:


ngap/handler.go
```diff
@@ -635,9 +637,15 @@ func HandleNGSetupRequest(ran *context.AmfRan, message *ngapType.NGAPPDU) {
 
 	if cause.Present == ngapType.CausePresentNothing {
 		ngap_message.SendNGSetupResponse(ran)
+		//send nf(gnb) status notification
+		gnbStatus := mi.MetricEvent{EventType: mi.CNfStatusEvt,
```
thakurajayL, line 641: Isn't it good to have this gNB state in sctplb ?

service/init.go
```diff
@@ -114,6 +116,11 @@ func (amf *AMF) Initialize(c *cli.Context) error {
 
 	amf.setLogLevel()
 
+	//Initiating a server for profiling
+	go func() {
+		http.ListenAndServe(":5001", nil)
```
thakurajayL, line 121: worried about hardcoding.  See example of gnbsim.. I believe we pass profilePort as arg.

metrics/kafka.go
```diff
@@ -0,0 +1,109 @@
...
+	return nil
+}
+
+/*
+var amfId string
+
+func PublishMsgEvent(msgType mi.AmfMsgType) error {
+
+	if amfId == "" {
+		amfId = os.Getenv("AMF_ID")
```
thakurajayL, line 87: Worried about env use like this.. Does not sound like good practice.

thakurajayL (changes requested):

metrics/kafka.go
```diff
@@ -0,0 +1,109 @@
...
+var StatWriter Writer
+
+func InitialiseKafkaStream(config *factory.Configuration) error {
+
+	brokerUrl := "sd-core-kafka-headless:9092"
+	topicName := "sdcore-data-source-amf"
+
+	if config.KafkaInfo.BrokerUri != "" && config.KafkaInfo.BrokerPort != 0 {
+		brokerUrl = fmt.Sprintf("%s:%d", config.KafkaInfo.BrokerUri, config.KafkaInfo.BrokerPort)
+		logger.KafkaLog.Debugf("initialise kafka broker url [%v]", brokerUrl)
```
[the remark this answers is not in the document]
UpadhyayaAnkur: Agree, done

service/init.go
UpadhyayaAnkur: Shall make it config based

metrics/kafka.go
UpadhyayaAnkur: Right, its expensive, shall replace it

ngap/handler.go
UpadhyayaAnkur: @vthiruveedula : Lets discuss this

thakurajayL (approved):

PR #138, opened 2023-11-23
fix: fetches nssf uri even if already cached

### LINKED ISSUE #133: AMF uses old NSSF address after NSSF IP changed
## Description

When the NSSF IP addresses changes (ex. after a pod reschedule), the AMF will not use the new address and keep on using the old one resulting in the following error:

```
2023-11-22T18:50:37.424Z [amf] 2023-11-22T18:50:37Z [ERRO][AMF][GMM][AMF_UE_NGAP_ID:13864935][SUPI:imsi-208930100007487][SUCI:suci-0-208-93-0-0-0-0100007487] Select SMF failed: NSSelection Get Error[NSSF No Response]
```

The issue seems to be caused by the fact that the AMF only gets the NSSF address from the NRF once when it starts up and keeps on using this address forever. Reference here:
- https://github.com/omec-project/amf/blob/master/consumer/sm_context.go#L96

gruyaume:
## Description

This PR makes it so that  the `SelectSmf` function retrieves `SearchNssfNSSelectionInstance` even if there already is a stored value for the NSSF URI. This fixes an issue where the NSSF uri was only stored once on AMF startup and never retrieved again. This was a problem in cases where the NSSF crashed and its IP changed.

Fixes #133

### DIFF
--- a/consumer/sm_context.go
+++ b/consumer/sm_context.go
@@ -92,16 +92,14 @@ func SelectSmf(
 
 	nsiInformation := ue.GetNsiInformationFromSnssai(anType, snssai)
 	if nsiInformation == nil {
-		if ue.NssfUri == "" {
-			// TODO: Set a timeout of NSSF Selection or will starvation here
-			for {
-				if err := SearchNssfNSSelectionInstance(ue, nrfUri, models.NfType_NSSF,
-					models.NfType_AMF, nil); err != nil {
-					ue.GmmLog.Errorf("AMF can not select an NSSF Instance by NRF[Error: %+v]", err)
-					time.Sleep(2 * time.Second)
-				} else {
-					break
-				}
+		// TODO: Set a timeout of NSSF Selection or will starvation here
+		for {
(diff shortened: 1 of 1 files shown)


consumer/sm_context.go
```diff
@@ -92,16 +92,14 @@ func SelectSmf(
...
-				if err := SearchNssfNSSelectionInstance(ue, nrfUri, models.NfType_NSSF,
-					models.NfType_AMF, nil); err != nil {
-					ue.GmmLog.Errorf("AMF can not select an NSSF Instance by NRF[Error: %+v]", err)
-					time.Sleep(2 * time.Second)
-				} else {
-					break
-				}
+		// TODO: Set a timeout of NSSF Selection or will starvation here
+		for {
+			if err := SearchNssfNSSelectionInstance(ue, nrfUri, models.NfType_NSSF,
```
thakurajayL, line 97: Something to be done if not already available. NRF results are cached internally. We should see if we can cache this result as well. that code is part of NRF repo. It helps in performance. Also upon any failure in NRF cache should be refreshed.

thakurajayL (approved):

PR #150, opened 2023-11-27
fix: Reenable nakedret linter and fix found issues

ghislainbourgeois:
Re-enables the `nakedret` linter and fix the issues found by it. This PR only makes the current implicit behaviour explicit to satisfy the linter. It does not change any current behaviors.


gab-arrobo (approved):
+1

PR #156, opened 2023-12-05
chore: Simplifies CI workflows

gruyaume:
## Description

This PR simplifies the Github Actions CI workflow by combining all jobs that should be run on master and on PR's to master in one file. The only real change in behavior is that we no longer update an existing PR with the outcome of a `go fmt`. Since we run go lint, the CI will simply fail if a given PR has bad formatting.

Effectively, this reduces the amount of CI lines of code by more than 50% without sacrificing anything.

## Rationale

It was hard to wrap one's head around what jobs were done at which moment because of the various files, their different conditions and the amount of commented workflow line. While the file representation was hard to read, the actual workflow was quite simple: Run X, Y, Z jobs on push to master and PR's to master. This PR here reflects this. 

## Screenshots

This cleans up the way checks are represented in Github actions. 

### Before
image

### After
image


.github/workflows/master.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Build
+        run: go build
+
+  docker-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build and push Docker image
```
gab-arrobo, line 32: This only `builds` the Docker image, but it does not `push` it. Should we also add `docker-push`? Should `docker-push` be applicable only for `push` to master? Any thoughts on this? Moreover, are you planning to do this same thing for all other repos (including the UPF)?

.github/workflows/master.yml
```diff
@@ -0,0 +1,69 @@
...
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v4
```
gab-arrobo, line 20: I just tested v5 and it worked fine

```suggestion
      - uses: actions/setup-go@v5
```

.github/workflows/master.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Build and push Docker image
+        run: make docker-build
+
+  lint:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v4
```
gab-arrobo, line 41: I just tested v5 and it worked fine

```suggestion
      - uses: actions/setup-go@v5
```

.github/workflows/master.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: reuse lint
+        run: make check-reuse
+
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v4
```
gab-arrobo, line 64: I just tested v5 and it worked fine

```suggestion
      - uses: actions/setup-go@v5
```

.github/workflows/master.yml
gruyaume: The current CI does not `docker push` anywhere and this change here keeps this behaviour. We may add it in the future but this was not the purpose of this change.

And yes the goal is to improve consistency between workflows of the different NF's. The workflow may end up being identical for all of them but it can be much closer than what it is right now.

gruyaume: I changed the description to only contain "Build Docker image"

.github/workflows/master.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          go-version: 1.21
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3.7.0
+        with:
+          version: latest
+          args: -v --config ./.golangci.yml
+
+  license-check:
```
gab-arrobo, line 51: This GHA does exactly the same as the Jenkins' License Verification. So, I think the one from Jenkins should be removed. Let's bring this up in the next Aether TST meeting.

gab-arrobo (commented):
+1

gab-arrobo (approved):
+1 (Tested changes locally)

PR #165, opened 2023-12-13
chore: Re-enable errcheck linter with required fixes

ghislainbourgeois:
Enables the `errcheck` linter and add error checking where missing, simply logging the error.

Also remove the `deadcode` and `varcheck` linters as they are abandoned by their upstream.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -247,7 +247,6 @@ linters:
     # - structcheck
     - varcheck
     - ineffassign
-    # - deadcode
     - typecheck
     # Additional
     # - lll
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -245,7 +245,6 @@ linters:
     # - unused
     # - gosimple
     # - structcheck
-    - varcheck
     - ineffassign
     - typecheck
     # Additional
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -240,7 +240,7 @@ linters:
   enable:
     - gofmt
     # - govet
-    # - errcheck
+    - errcheck
     # - staticcheck
     # - unused
     # - gosimple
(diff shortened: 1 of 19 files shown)


gab-arrobo:
ok to test

.golangci.yml
```diff
@@ -240,14 +240,12 @@ linters:
   enable:
     - gofmt
     # - govet
-    # - errcheck
+    - errcheck
     # - staticcheck
     # - unused
     # - gosimple
     # - structcheck
-    - varcheck
```
gab-arrobo, line 248: Why is this one deleted?

.golangci.yml
```diff
@@ -240,14 +240,12 @@ linters:
...
     - ineffassign
-    # - deadcode
```
gab-arrobo, line 250: Why is this one deleted? Do we want to reenable it at some point in the future? FYI, when enabled, this check fails because there are two functions (`includeConfiguredNssaiCheck` and `refreshSmfProfiles`) that are unused.

.golangci.yml
ghislainbourgeois: When running, there is a warning that the linter is abandonned by its upstream. The link is also broken on the list, returning a 404, so I do not think we can really rely on that one: https://golangci-lint.run/usage/linters/

.golangci.yml
ghislainbourgeois: This linter has also been abandonned by its upstream, last seeing an update 7 years ago: https://github.com/remyoudompheng/go-misc/tree/master/deadcode.

The `unused` linter will probably cover the same use cases and is maintained.

gab-arrobo (approved):
+1 (tested changes locally)

PR #171, opened 2024-01-03
chore: Uses go version from go.mod

gruyaume:
# Description

We use the go version from go.mod instead of specifying it in CI. This makes it so that when we will bump the Go version we will have to do it in less places and therefore avoids potential differences between the go version running in CI and the one in go.mod.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version-file: "go.mod"
+          go-version-file: 'go.mod'
 
       - name: Build
         run: go build
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.21
+          go-version-file: "go.mod"
 
       - name: Build
         run: go build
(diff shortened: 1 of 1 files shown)


gab-arrobo:
ok to test

.github/workflows/master.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.21
+          go-version-file: "go.mod"
```
gab-arrobo, line 22: As per the [setup-go](https://github.com/marketplace/actions/setup-go-environment#getting-go-version-from-the-gomod-file)
```suggestion
          go-version-file: 'go.mod'
```

.github/workflows/master.yml
```diff
@@ -40,7 +40,7 @@ jobs:
...
```
gab-arrobo, line 43: As per the [setup-go](https://github.com/marketplace/actions/setup-go-environment#getting-go-version-from-the-gomod-file)
```suggestion
          go-version-file: 'go.mod'
```

.github/workflows/master.yml
```diff
@@ -63,7 +63,7 @@ jobs:
...
```
gab-arrobo, line 66: As per the [setup-go](https://github.com/marketplace/actions/setup-go-environment#getting-go-version-from-the-gomod-file)
```suggestion
          go-version-file: 'go.mod'
```

gab-arrobo (approved):
+1

PR #186, opened 2024-02-01
chore: Move fsm dependency to util project

ghislainbourgeois:
This PR moves the import path of the `fsm` module to use the module under `util`. This was tested with a successful gnbsim simulation.


gab-arrobo (approved):
+1 (tested changes with AiaB)

PR #187, opened 2024-02-01
chore: Move idgenerator dep to util project

ghislainbourgeois:
Changes the import path of the `idgenerator` module to use the `util` project. Tested with a successful simulation with gnbsim.

### DIFF
--- a/context/amf_ue.go
+++ b/context/amf_ue.go
@@ -24,7 +24,6 @@ import (
 	"github.com/omec-project/amf/logger"
 	"github.com/omec-project/amf/metrics"
 	"github.com/omec-project/amf/protos/sdcoreAmfServer"
-	"github.com/omec-project/idgenerator"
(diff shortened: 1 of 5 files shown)


gab-arrobo:
@ghislainbourgeois, just to confirm, you are verifying that the code (functions, variables, etc.) from the "tiny" repo (`idgenerator` in this case) is exactly the same as the one in the `util` repo (`util/idgenerator`), correct?

ghislainbourgeois:
> @ghislainbourgeois, just to confirm, you are verifying that the code (functions, variables, etc.) from the "tiny" repo (`idgenerator` in this case) is exactly the same as the one in the `util` repo (`util/idgenerator`), correct?

There are small differences that should not impact the behavior. But I ran a full test to make sure.

gab-arrobo (approved):
+1 (tested changes with AiaB)

PR #207, opened 2024-03-13
 Parameterise webui url

gatici:
The Webui url is hardcoded in network function services. This PR makes Webui url changeable having an entry in the configuration file.


gatici:
The change is tested and it passed the simulation successfully.

gab-arrobo:
> The change is tested and it passed the simulation successfully.

Hi @gatici, have you tested the changes using/setting a different port number?

gatici:
> > The change is tested and it passed the simulation successfully.
> 
> Hi @gatici, have you tested the changes using/setting a different port number?

Hello @gab-arrobo 

Port number 9876 is hard-coded in the webui. Unless we try the port in webui, it will fail if change to an another port number.

gab-arrobo (approved):
+1

PR #249, opened 2024-05-13
fix: create collections only when database is enabled

dariofaccin:
This PR fixes the wrong behaviour for which the AMF tries to create collections mongoDB even when the `EnableDbStore` configuration option is set to false.

### DIFF
--- a/service/init.go
+++ b/service/init.go
@@ -358,7 +358,7 @@ func (amf *AMF) Start() {
 	}
 
 	if self.EnableDbStore {
-		go self.SetupAmfCollection()
+		go context.SetupAmfCollection()
 	}
 
 	signalChannel := make(chan os.Signal, 1)
--- a/service/init.go
+++ b/service/init.go
@@ -357,7 +357,9 @@ func (amf *AMF) Start() {
 		go StartGrpcServer(self.SctpGrpcPort)
 	}
 
-	go context.SetupAmfCollection()
+	if self.EnableDbStore {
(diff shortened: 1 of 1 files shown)


gab-arrobo:
@gatici, what are your thoughts on this PR?

gab-arrobo:
@dariofaccin please address the GHAs' issues. Thanks!

gatici:
> @gatici, what are your thoughts on this PR?

If the `EnableDbStore` is set to False, DB connection will not set up and the collections will not be created. However, the current code still tries to connect DB and create collections. Those part of code should run according to the value of  `EnableDbStore` configuration attribute. I support the purpose of this change.

dariofaccin:
@gab-arrobo could you please retrigger the CI workflow? Thanks

gab-arrobo (approved):
+1 (tested changes with OnRamp)

service/init.go
```diff
@@ -357,7 +357,9 @@ func (amf *AMF) Start() {
 		go StartGrpcServer(self.SctpGrpcPort)
 	}
 
-	go context.SetupAmfCollection()
+	if self.EnableDbStore {
+		go context.SetupAmfCollection()
+	}
```
thakurajayL, line 362: Thanks for change. I see in some cases EnableDbStore  check is done inside the function and sometimes its done outside. But your change is perfect. Thanks for your contribution.

thakurajayL (approved):

Issue #262, opened 2024-06-13
Kafka errors when not using Kafka


gruyaume:
## Description

In our deployment of the AMF we are not using Kafka, therefore I would expect the AMF not to try to write/read anything from Kafka. We do see errors of the kind copied below in the logs.

## Logs

```
2024-06-13T19:18:00.967Z [amf] 2024-06-13T19:18:00Z [ERRO][AMF][Kafka] Error publishing nf status event: dial tcp: lookup sd-core-kafka-headless on 10.152.183.10:53: no such host
2024-06-13T19:18:00.967Z [amf] 2024-06-13T19:18:00Z [ERRO][AMF][Kafka] Could not publish ue context event, error [dial tcp: lookup sd-core-kafka-headless on 10.152.183.10:53: no such host]
```

thakurajayL:
Can you please raise pr with moving InitialiseKafkaStream conditional and execute it only if kafka configuration exist?

gruyaume:
I don't mind adding this to our backlog, this will likely not be top priority however. I opened it here as it's closer to the source.

thakurajayL:
https://github.com/omec-project/amf/pull/263 this PR should suppirt kafka disable.

thakurajayL:
Fixed. Next release should have the fix available.

PR #263, opened 2024-06-15
Support to disable Kafka

thakurajayL:


gab-arrobo:
close #262

context/amf_ran.go
```diff
@@ -73,8 +73,10 @@ func (ran *AmfRan) Remove() {
 			NfStatus: mi.NfStatusDisconnected, NfName: ran.GnbId,
 		},
 	}
-	if err := metrics.StatWriter.PublishNfStatusEvent(gnbStatus); err != nil {
-		ran.Log.Errorf("Could not publish NfStatusEvent: %v", err)
+	if metrics.Status() {
+		if err := metrics.StatWriter.PublishNfStatusEvent(gnbStatus); err != nil {
+			ran.Log.Errorf("Could not publish NfStatusEvent: %v", err)
+		}
```
gab-arrobo, line 79: I see you are adding `if` statements (`if metrics.Status() {`) in the different places where `metrics.StatWriter.PublishNfStatusEvent(gnbStatus)` is called. What about adding a single `if` statement inside `PublishNfStatusEvent(msgEvent mi.MetricEvent)` such that there are changes in fewer number of files and looks cleaner?

Something like this in `metrics/kafka.go`:
```go
func (writer Writer) PublishNfStatusEvent(msgEvent mi.MetricEvent) error {
	if !metrics.Status() {
		logger.KafkaLog.Info("Kafka is disabled")
		return nil
	}
	if msg, err := json.Marshal(msgEvent); err != nil {
		return err
	} else {
		logger.KafkaLog.Debugf("publishing nf status event[%s] ", msg)
		if err := StatWriter.SendMessage(msg); err != nil {
			logger.KafkaLog.Errorf("Error publishing nf status event: %v", err)
		}
	}
	return nil
}
```

gab-arrobo (commented):
To be able to effectively disable publishing of data by the `amf` and `smf`, changes need to be done in the helm charts. Does it mean that we need to expose (override) these parameters/values from the sdcore-5g-values.yaml file? Currently, it only exposes the option to deploy (or not) `kafka` pod regardless of what is configured in the `amf` and/or `smf`. Does any other NF use kafka? If not, what about "controlling" the deployment of `kafka` if and only if the `amf` or `smf` needs to publish topics/data?

context/amf_ran.go
thakurajayL: I thought about this but it also means we are initializing Kafka code (Since to come here we need writer). I wanted to avoid that.

thakurajayL: I can do the check in metric in case of - PublishNfStatusEvent but similar check is not good for PublishUeCtxtEvent

gab-arrobo: Got it.

gab-arrobo (approved):
+1

gab-arrobo:
> To be able to effectively disable publishing of data by the `amf` and `smf`, changes need to be done in the helm charts. Does it mean that we need to expose (override) these parameters/values from the sdcore-5g-values.yaml file? Currently, it only exposes the option to deploy (or not) `kafka` pod regardless of what is configured in the `amf` and/or `smf`. Does any other NF use kafka? If not, what about "controlling" the deployment of `kafka` if and only if the `amf` or `smf` needs to publish topics/data?

@thakurajayL, are you planning to make changes to helm charts and sdcore-5g-values.yaml to expose the option to effectively  disable kafka inside the `amf` and `smf`?

thakurajayL:
> > To be able to effectively disable publishing of data by the `amf` and `smf`, changes need to be done in the helm charts. Does it mean that we need to expose (override) these parameters/values from the sdcore-5g-values.yaml file? Currently, it only exposes the option to deploy (or not) `kafka` pod regardless of what is configured in the `amf` and/or `smf`. Does any other NF use kafka? If not, what about "controlling" the deployment of `kafka` if and only if the `amf` or `smf` needs to publish topics/data?
> 
> @thakurajayL, are you planning to make changes to helm charts and sdcore-5g-values.yaml to expose the option to effectively disable kafka inside the `amf` and `smf`?

Its already there,
diff --git a/roles/core/templates/sdcore-5g-values.yaml b/roles/core/templates/sdcore-5g-values.yaml
index 485448c..86cfbbc 100644
--- a/roles/core/templates/sdcore-5g-values.yaml
+++ b/roles/core/templates/sdcore-5g-values.yaml
@@ -73,6 +73,8 @@ omec-control-plane:
         amfcfg.conf:
           configuration:
             enableDBStore: true # Store AMF subscribers in the datastore
+            kafkaInfo:
+              brokerUri: ""

     # SMF config override. Refer to Helm Charts values for more options
     smf:
ajayonf@node0:~/aether-onramp/deps/5gc$

Let me know what else you are looking for.

gab-arrobo:
> > > To be able to effectively disable publishing of data by the `amf` and `smf`, changes need to be done in the helm charts. Does it mean that we need to expose (override) these parameters/values from the sdcore-5g-values.yaml file? Currently, it only exposes the option to deploy (or not) `kafka` pod regardless of what is configured in the `amf` and/or `smf`. Does any other NF use kafka? If not, what about "controlling" the deployment of `kafka` if and only if the `amf` or `smf` needs to publish topics/data?
> > 
> > 
> > @thakurajayL, are you planning to make changes to helm charts and sdcore-5g-values.yaml to expose the option to effectively disable kafka inside the `amf` and `smf`?
> 
> Its already there, diff --git a/roles/core/templates/sdcore-5g-values.yaml b/roles/core/templates/sdcore-5g-values.yaml index 485448c..86cfbbc 100644 --- a/roles/core/templates/sdcore-5g-values.yaml +++ b/roles/core/templates/sdcore-5g-values.yaml @@ -73,6 +73,8 @@ omec-control-plane: amfcfg.conf: configuration: enableDBStore: true # Store AMF subscribers in the datastore
> 
> * ```
>          kafkaInfo:
>   ```
> * ```
>            brokerUri: ""
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   # SMF config override. Refer to Helm Charts values for more options
>   smf:
>   ajayonf@node0:~/aether-onramp/deps/5gc$
> 
> Let me know what else you are looking for.

What I mean is the following two things:
1. I thonk we should we expose this option in `sdcore-5g-values.yaml` file in `OnRamp` because currently, we can "disable" kafka (not deploy the kafka pod) but we can not do any change related to kafka in the amf and smf (in sdcore-5g-values.yaml.
2. Currently we have two "parallel" options to "enable/disable" kafka in the helm charts. One is about the deployment of kafka (pod) and the other is about using kafka inside the amf and smf pods. So, I was wondering if you are planning to "unify" these two options such that the kafka pod is deployed only if the amf or smf has it enabled.

PR #264, opened 2024-06-16
Using enableKafka variable to control kafka

thakurajayL:


factory/factory.go
```diff
@@ -34,6 +34,10 @@ func InitConfigFactory(f string) error {
 		if AmfConfig.Configuration.WebuiUri == "" {
 			AmfConfig.Configuration.WebuiUri = "webui:9876"
 		}
+		if AmfConfig.Configuration.KafkaInfo.EnableKafka == nil {
```
gab-arrobo, line 37: Would not this mean that this will always be true (i.e., not `nil`) if the `enableKafka` field is present (regardless of its value)? or am I overlooking something?

gab-arrobo (approved):
+1

PR #291, opened 2024-08-30
AMF pod crash when Mobile data disabled and re-enabled in UE

ashithacdac:
Issue: We have successfully connected the SD core with the OAI-gNB (usrp B210) and real UE. However, the pod crashes consistently when we disable and re-enable mobile data.

Troubleshoot: We identified the root cause of the pod crash: the Pdusessionresourcesetupresponse message contains both PDUSessionResourceSetupListSures and PDUSessionResourceFailedToSetupListSures IE's. While PDUSessionResourceSetupListSures contains the expected pduSessionId matching the one sent in the PDUSessionResourceSetupRequest, PDUSessionResourceFailedToSetupListSures contains a pduSessionId of 0, causing the pod to crash because the corresponding SmContext cannot be found.

Fix: Added continue if SmContext is not found.


ak2btvm (approved):
All suggested changes by @thakurajayL  and @gab-arrobo are incorporated. Now it seems okay to merge.

thakurajayL (approved):

PR #305, opened 2024-09-23
N2 handover - Downlink RAN Status Transfer failure fix

JayaramRCDAC:
Issue:
N2 Handover failure due to SendDownlinkRanStatusTransfer() not being called after Uplink RAN Status Transfer.

Troubleshoot:
Upon investigation, it was found that the function SendDownlinkRanStatusTransfer() was not triggered, which led to the failure of the N2 handover process.

Fix:
The issue has been fixed by ensuring that SendDownlinkRanStatusTransfer() is called with the target UE after the Uplink RAN Status Transfer.

Supporting Data:
[N2-Handover-PCAP-Logs.zip](https://github.com/user-attachments/files/17095409/N2-Handover-PCAP-Logs.zip)


thakurajayL:
Will review it today !

ngap/handler.go
```diff
@@ -3801,6 +3801,20 @@ func HandleUplinkRanStatusTransfer(ran *context.AmfRan, message *ngapType.NGAPPD
...
+	// Send Downlink RAN Status Transfer
+	targetUe := ranUe.TargetUe
+	if targetUe != nil {
+		logger.NgapLog.Info("Target AMF UE NGAP Id: ", targetUe.AmfUeNgapId)
+		logger.NgapLog.Info("Target RAN UE NGAP Id: ", targetUe.RanUeNgapId)
+	}
+
+	if rANStatusTransferTransparentContainer != nil {
+		ngap_message.SendDownlinkRanStatusTransfer(amfUe.RanUe[models.AccessType__3_GPP_ACCESS], *rANStatusTransferTransparentContainer)
+		ngap_message.SendDownlinkRanStatusTransfer(targetUe, *rANStatusTransferTransparentContainer)
```
gab-arrobo, line 3814: What if `targetUe` is `nil`? Are not you checking that because it is getting checked in `SendDownlinkRanStatusTransfer()`? or why are not you checking it?

gab-arrobo (commented):
How can I replicate the issue such that I can test your PR?

ngap/handler.go
JayaramRCDAC: Yes, I didn't add a nil check because the SendDownlinkRanStatusTransfer function already handles it for ue (i.e., targetUe). However, I can add an additional nil check if we want to log an early message or perform other actions when targetUe is nil. Let me know if you think logging or any specific action should be done in that case.

gab-arrobo: No need to add another check. Thr main thing was that I wanted to double check I understood the reason for that.

gab-arrobo:
How can I replicate the issue such that I can test the PR?

JayaramRCDAC:
> How can I replicate the issue such that I can test your PR?

We have tested N2 handover using both a simulation tool and a physical UE. In both cases, the gNodeB did not receive the Downlink RAN Status Transfer. 
If you'd like, we can share the details of the test cases to help you replicate the issue.

ngap/handler.go
thakurajayL, line 3814: Why this call is made 2 times. Could you please help with specific 23.502 reference. As per my reading 23.502 
4.9.1.3.3 Execution phase step 2c shows message from target AMF to target gNB

JayaramRCDAC: I accidentally added the first function call. As above reference, only the message from the target AMF to the target gNB is required. 
So, only `ngap_message.SendDownlinkRanStatusTransfer(targetUe, *rANStatusTransferTransparentContainer)` should be there.

thakurajayL (approved):

PR #307, opened 2024-09-25
Authentication Failure - ngKSI already in use

ashithacdac:
Issue: 
Authentication failure with cause ngKSI already in use

Troubleshoot:
Identified the root cause: the core includes the same ngKSI value from the Registration Request
 in the Authentication request

Fix: 
Updated the ngKSI value after the 5G AKA procedure

Supporting Data:
[Authentication_fail_ngksi_already_in_use.zip](https://github.com/user-attachments/files/17126129/Authentication_fail_ngksi_already_in_use.zip)


gmm/handler.go
```diff
@@ -1558,6 +1558,16 @@ func AuthenticationProcedure(ue *context.AmfUe, accessType models.AccessType) (b
 	ue.AuthenticationCtx = response
 	ue.ABBA = []uint8{0x00, 0x00} // set ABBA value as described at TS 33.501 Annex A.7.1
 
+	if ue.NgKsi.Tsc == models.ScType_NATIVE && ue.NgKsi.Ksi != 7 {
+		// As per the Specification 24.501 - 5.4.1.3.2 Authentication initiation by the network
+		if ue.NgKsi.Ksi < 6 { // ksi is range from 0 to 6
```
thakurajayL, line 1563: ue.NgKsi.Ksi = (ue.NgKsi.Ksi +1) % 7

ashithacdac: Incorporated the suggested changes.

gmm/handler.go
```diff
@@ -1558,6 +1558,12 @@ func AuthenticationProcedure(ue *context.AmfUe, accessType models.AccessType) (b
 	ue.AuthenticationCtx = response
 	ue.ABBA = []uint8{0x00, 0x00} // set ABBA value as described at TS 33.501 Annex A.7.1
 
+	if ue.NgKsi.Tsc == models.ScType_NATIVE && ue.NgKsi.Ksi != 7 {
```
thakurajayL, line 1561:
```suggestion
	if ue.NgKsi.Tsc == models.ScType_NATIVE  {
```

thakurajayL: ue.NgKsi.Ksi will never be 7.

gmm/handler.go
```diff
@@ -1558,6 +1558,12 @@ func AuthenticationProcedure(ue *context.AmfUe, accessType models.AccessType) (b
...
+		// As per the Specification 24.501 - 5.4.1.3.2 Authentication initiation by the network
```
gab-arrobo, line 1562: BTW, should not 33.501 the correct document to reference?
```suggestion
		// As per the Specification 33.501 - 6.2.3.2 Key identification
```

gmm/handler.go
ashithacdac: I would appreciate clarification on the statement regarding Key lifetimes in the specification 33.501 section 6.2.3.3 which states 'In case the UE does not have a valid KAMF, an ngKSI with value "111" shall be sent by the UE to the network, which can initiate (re)authentication procedure to get a new KAMF based on a successful primary authentication.'

gab-arrobo: > I would appreciate clarification on the statement regarding Key lifetimes in the specification 33.501 section 6.2.3.3 which states 'In case the UE does not have a valid KAMF, an ngKSI with value "111" shall be sent by the UE to the network, which can initiate (re)authentication procedure to get a new KAMF based on a successful primary authentication.'

@thakurajayL what are yout thoughts about this?

thakurajayL: I read the section you mentioned. It just says that if UE sends 111 then it just means UE does not have Kamf  and its requesting new keys. In my opinion its perfectly fine to remove the != 7 check in the above code and let the keys rotate from 0 to 6.

gab-arrobo:
@ashithacdac any update about this?

gmm/handler.go
ashithacdac: Incorporated the suggested changes.

ashithacdac:
@gab-arrobo Incorporated the suggested changes.

gab-arrobo:
@ashithacdac, FYI, it looks like your PR was not properly rebased

PR #313, opened 2024-10-01
Number of fixes in AMF to handle multiple service req

thakurajayL:
1. UE Context Release handling leads to total 3 messages. Final message is UE Context Release Complete is sent by gNB. Service request was reaching AMF before UE Context Release Complete and this leads to crash, since UE to RAN-UE linking gets removed while processing Complete message. Added necessary protection to fix the crash

2. Other fixes involve AMF should not handle the new calls if it has not yet received any config else AMF crashes due to no config


gab-arrobo (approved):
+1 (Tested changes with OnRamp)

PR #330, opened 2024-10-21
fix: observe GRPC client

gatici:
This PR aims to solve GPRC client connection issues which appears when GPRC server (webconsole) is restarted.

If the Webconsole is restarted, other modules loses the GRPC connection to Webconsole
They tries to connect again for hours with no success.

2024-09-26T11:32:10.014Z [smf] 2024-09-26T11:32:10Z [ERRO][Config5g][GRPC] Connectivity status idle, trying to connect again

NF observes the GRPC client availability and connection status in an infinite loop.
If the connection is ready it does nothing. If GRPC client does not exist, it creates new client. If GRPC client connection is not ready, it closes the existing client and creates another client.
Only one GRPC client exists for any time slot.


gab-arrobo (commented):
I see that changes are not consistent across the different NFs (in some NFs the `updateConfig` is in `service/init.go` like this in NF/PR and in other NFs, the same function goes in `factory/factory.go`). Would this be a good time to make the change consistent across all NFs?

service/init.go
```diff
@@ -144,10 +144,8 @@ func (amf *AMF) Initialize(c *cli.Context) error {
 		factory.AmfConfig.Configuration.ServedGumaiList = nil
 		factory.AmfConfig.Configuration.SupportTAIList = nil
 		factory.AmfConfig.Configuration.PlmnSupportList = nil
-		logger.InitLog.Infoln("reading Amf related configuration from ROC")
-		client := gClient.ConnectToConfigServer(factory.AmfConfig.Configuration.WebuiUri)
-		configChannel := client.PublishOnConfigChange(true)
-		go amf.UpdateConfig(configChannel)
+		logger.InitLog.Infoln("Reading Amf related configuration from ROC")
```
gab-arrobo, line 147:
```suggestion
		logger.InitLog.Infoln("reading Amf related configuration from ROC")
```

gab-arrobo (approved):

PR #335, opened 2024-10-31
Refactor path for configuration file, opton(s) and Docker image

gab-arrobo:


Dockerfile
```diff
@@ -27,6 +25,7 @@ RUN make all
 
 FROM alpine:3.20 AS amf
 
+LABEL maintainer="Aether SD-Core <dev@lists.aetherproject.org>"
```
thakurajayL, line 28: Could you pls send test mail on this.

factory/factory.go
```diff
@@ -43,7 +43,7 @@ func InitConfigFactory(f string) error {
 	return nil
 }
 
-func UpdateAmfConfig(f string) error {
+func UpdateConfig(f string) error {
```
thakurajayL, line 46: Though I agree with this change there are plenty of variables with name amfConfig... But I do not want those many variable changes.

gab-arrobo: We can make those changes in another PR such that a single does not become to big to review. What do you think?

Dockerfile
gab-arrobo: Done. I just sent an email there. One caveat is that you need to subscribe to send/receive emails

thakurajayL: Is that information captured somewhere? How to subscriber for mailer?

factory/factory.go
thakurajayL: Changing function name is enough. Sometimes its easy to keep amf in the variable name. I would not recommend that change.

thakurajayL (approved):

Dockerfile
gab-arrobo: I do not see that being documented anywhere. As a matter of fact, I just signed up to the mailing list

PR #347, opened 2024-11-11
Update GO version due to gRPC-Go's prerequisites

gab-arrobo:
From prerequisites subsection (https://pkg.go.dev/google.golang.org/grpc@v1.68.0):
- Go: any one of the two latest major releases (i.e., 1.22 or 1.23)


amfTest/amfcfg.yaml
```diff
@@ -102,22 +102,13 @@ configuration:
 logger:
   AMF:
     debugLevel: info
-    ReportCaller: false
   NAS:
     debugLevel: info
-    ReportCaller: false
```
thakurajayL, line 108: We are removing reportCaller because new logging library does not need this right?

amfTest/amfcfg.yaml
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2021 Open Networking Foundation <info@opennetworking.org>
 #
 # SPDX-License-Identifier: Apache-2.0
-#
```
thakurajayL, line 4: add copyright in this file as well.

.golangci.yml
```diff
@@ -233,7 +234,7 @@ linters:
     # Additional
     # - lll
     - godox
-    #- gomnd
+    # - mnd
```
thakurajayL, line 237: still remains commented. What is this change.

amfTest/amfcfg.yaml
gab-arrobo: Correct

amfTest/amfcfg.yaml
gab-arrobo: Done

.golangci.yml
gab-arrobo: The `gomnd` (magic number) linter does not exist anymore. Now, its name is `mnd`. This linter is about defining constant values/variables (e.g., const TIMEOUT = 2) and then use the `TIMEOUT` in the code. This is something that needs to be done but I cannot do it at the moment and should be done in a future PR.

PR #348, opened 2024-11-12
SMF Selection Fail Bug Fix

ashithacdac:
Issue:
SMF Selection Failure when connecting UEs to Core, especially for already connected UEs.

Troubleshoot:
Troubleshooting revealed that failure occurs when UE sends nil or 1 'sd value' in Security Mode Complete message.

Fix:
1. Modified InSubscribedNssai() to consider only 'sst value'.
2. Set targetSNssai.Sd to SubscribedSnssai.Sd, ensuring consistent handling.

Supporting Data:
[SMF Selection Fail.zip](https://github.com/user-attachments/files/17716255/SMF.Selection.Fail.zip)


thakurajayL:
Would be good if you can provide reference to below things

1. For provided pcap - could you pls provide reference to packet numbers so its easy to quickly understand the issue
2. Could you please provide some 3gpp reference to this change?
3. Does this happen with tool or real UE?
4. Do you mean UE connects, de-registers and then trying to connect one more time and it sees failures?
Thanks

ashithacdac:
1.Please refer to packets 14-23.
2.3GPP specification 24.501, section 4.6, "Network Slicing," regarding optional SD.
3.This issue happens with a real UE.
4.Yes

thakurajayL (approved):

PR #358, opened 2024-12-04
fix: check if a message read from channel to prevent nil pointer dereference

gatici:
Aims to fix https://github.com/canonical/sdcore-amf-k8s-operator/issues/472

Invalid memory address or nil pointer dereference error occurs when producer.HandleRegistrationStatusUpdateRequest is called
(producer/ue_context.go:599).  This PR first checks if a message is read through the channel before proceeding with response data and problem details.


gab-arrobo (approved):
+1

PR #402, opened 2025-02-15
Fix AMF crash

thakurajayL:
AMF sent Initial Context Setup message to gNB, but before gNB initiated context release while context setup was in-progress. This leads to PDU session release and later when Initial Context Setup failure is received from the gNodeB, amf failed to find pdu-session.

its wise to just process initial context setup failure message at AMF & skip processing which sends message towards SMF.


gab-arrobo (approved):
+1

PR #440, opened 2025-04-11
chore: enable static check and govet in the CI

patriciareinoso:
this PR enables `staticcheck` and `govet` in the CI

- fix a variable declaration
- remove explicit call to embedded structs
- replace if with switch

govet's `fieldaligment` remains disabled


gab-arrobo (approved):
+1

PR #475, opened 2025-06-30
chore: run E2E tests automatically

gatici:
Run E2E tests for pull requests as part of CI process. Hence, we make sure that integrations tests will pass without manual testing.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,6 +51,7 @@ jobs:
       branch_name: ${{ github.ref }}
 
   e2e-tests:
+    if: github.event_name == 'pull_request'
     uses: omec-project/.github/.github/workflows/e2e-test.yml@main
     with:
       branch_name: ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -49,3 +49,8 @@ jobs:
     uses: omec-project/.github/.github/workflows/unit-test.yml@main
     with:
       branch_name: ${{ github.ref }}
+
+  e2e-tests:
+    uses: omec-project/.github/.github/workflows/e2e-test.yml@main
+    with:
+      branch_name: ${{ github.ref }}
```
gab-arrobo, line 56: Given that this GHA takes ~10 mins to complete, I think we should only run it `on.pull_request` and not run it `on.push`. Current default behavior is to run it in both (pull_request and push)

gab-arrobo (changes requested):

.github/workflows/main.yml
gatici: I have pushed to wrong branch. Now, this is fixed.

Issue #442, opened 2025-04-15
delete all network slices do not remove elements from factory.AmfConfig.Configuration.ServedGumaiList stored in the AMF


patriciareinoso:
## Description

When receiving the network configuration from webconsole's via the gRPC service. The AMF stores the PLMN ID (MCC + MNC)  and TACS in `factory.AmfConfig.Configuration.ServedGumaiList `. This s used in the NF's profile when registering to the NRF.

```
2025-04-15T10:59:23.052Z [amf] 2025-04-15T10:59:23.052Z INFO    client/gClient.go:193   stream message received {"component": "Config5g", "category": "GRPC"}
2025-04-15T10:59:23.058Z [amf] 2025-04-15T10:59:23.057Z INFO    client/gClient.go:201   first time config received RestartCounter:2619283598 NetworkSlice:{Name:"test1" Nssai:{Sst:"1" Sd:"010203"} DeviceGroup:{Name:"cows" IpDomainDetails:{Name:"pool1" DnnName:"internet" UePool:"172.250.1.0/16" DnsPrimary:"8.8.8.8" Mtu:1460 UeDnnQos:{TrafficClass:{Name:"platinum" Qci:8 Arp:6 Pdb:300 Pelr:6}}} Imsi:"208930100007487"} Site:{SiteName:"demo" Gnb:{Name:"testpcf-gnbsim-gnbsim" Tac:1} Plmn:{mcc:"208" mnc:"93"} Upf:{UpfName:"upf-external.testpcf.svc.cluster.local"}} AppFilters:{PccRuleBase:{FlowInfos:{FlowDesc:"permit out ip from any to assigned" TosTrafficClass:"IPV4" FlowDir:BIDIRECTIONAL} RuleId:"DefaultRule" Qos:{Var5qi:9 Arp:{PL:1 PC:MAY_PREEMPT PV:PREEMPTABLE}} Priority:255}}} ConfigUpdated:1 {"component": "Config5g", "category": "GRPC"}
[...]

PR #486, opened 2025-07-23
feat: implement polling for NF configuration

dariofaccin:
This PR implements the new NF configuration system.
The AMF will poll every 5 seconds the webconsole to fetch the latest Access-Mobility configuration.
It reaches the webconsole trough the /nfconfig/access-mobility endpoint.
There is an exponential retry mechanism in case of failure.
This PR fixes https://github.com/omec-project/amf/issues/442
After this PR, we do not need to set `MANAGED_BY_CONFIG_POD` environment variable anymore. Configuration is done via the webconsole.


consumer/nf_management.go
```diff
@@ -0,0 +1,313 @@
...
+	profile.AmfInfo = &amfInfo
+	plmnCopy := make([]models.PlmnId, len(accessAndMobilityConfig))
+	for _, accessAndMobilityData := range accessAndMobilityConfig {
+		nfPlmn := models.PlmnId{
+			Mcc: accessAndMobilityData.PlmnId.GetMcc(),
+			Mnc: accessAndMobilityData.PlmnId.GetMnc(),
+		}
+		plmnCopy = append(plmnCopy, nfPlmn)
+	}
+	profile.PlmnList = &plmnCopy
```
gatici, line 67: We only use PLMN ID to create NF profile. If NSSAI or TACS has been changed, this should not trigger NRF registration.

service/init.go
```diff
@@ -375,14 +294,80 @@ func (amf *AMF) Start() {
...
+		polling.StartPollingService(ctx, factory.AmfConfig.Configuration.WebuiUri, func(cfg []nfConfigApi.AccessAndMobility) {
+			select {
+			case registrationChan <- cfg:
+			default:
+				logger.PollConfigLog.Warn("registrationChan full, dropping config")
+			}
+			select {
+			case contextUpdateChan <- cfg:
+			default:
+				logger.PollConfigLog.Warn("contextUpdateChan full, dropping config")
```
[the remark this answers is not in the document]
dariofaccin: Now `StartPollingService` accepts two more parameters: `registrationChannel` and `contextUpdateChannel`.

polling/nf_configuration.go
```diff
@@ -0,0 +1,144 @@
...
+	oldPlmnList := []nfConfigApi.PlmnId{}
+	newPlmnList := []nfConfigApi.PlmnId{}
+	for _, plmnSnssaiTac := range oldAccessMobilityConfig {
+		oldPlmnList = append(oldPlmnList, plmnSnssaiTac.PlmnId)
+	}
+	for _, plmnSnssaiTac := range newAccessMobilityConfig {
+		newPlmnList = append(newPlmnList, plmnSnssaiTac.PlmnId)
+	}
+	return !reflect.DeepEqual(oldPlmnList, newPlmnList)
+}
```
gab-arrobo, line 144: If I am not mistaken, `minDuration` is the last function (in the file) in the other NFs
```suggestion
func plmnListIsChanged(oldAccessMobilityConfig, newAccessMobilityConfig []nfConfigApi.AccessAndMobility) bool {
	oldPlmnList := []nfConfigApi.PlmnId{}
	newPlmnList := []nfConfigApi.PlmnId{}
	for _, plmnSnssaiTac := range oldAccessMobilityConfig {
		oldPlmnList = append(oldPlmnList, plmnSnssaiTac.PlmnId)
	}
	for _, plmnSnssaiTac := range newAccessMobilityConfig {
		newPlmnList = append(newPlmnList, plmnSnssaiTac.PlmnId)
	}
	return !reflect.DeepEqual(oldPlmnList, newPlmnList)
}

func minDuration(a, b time.Duration) time.Duration {
	if a < b {
		return a
	}
	return b
}
```

polling/nf_configuration_test.go
```diff
@@ -0,0 +1,290 @@
...
+			Mcc: mcc,
+			Mnc: mnc,
+		},
+		Snssai: nfConfigApi.Snssai{
+			Sst: sstint,
+			Sd:  &sd,
+		},
+		Tacs: tacs,
+	}
+}
```
gab-arrobo, line 46: What about returning an error instead of panic here and then check is `err != nil` to call t.Fatalf()`
```suggestion
func makeAccessMobilityConfig(mcc, mnc, sst string, sd string, tacs []string) (nfConfigApi.AccessAndMobility, error) {
	sstUint64, err := strconv.ParseUint(sst, 10, 8)
	if err != nil {
		return nfConfigApi.AccessAndMobility{}, err
	}
	sstint := int32(sstUint64)
	return nfConfigApi.AccessAndMobility{
		PlmnId: nfConfigApi.PlmnId{
			Mcc: mcc,
			Mnc: mnc,
		},
		Snssai: nfConfigApi.Snssai{
			Sst: sstint,
			Sd:  &sd,
		},
		Tacs: tacs,
	}, nil
}
```

```go
	accessMobility1, err := makeAccessMobilityConfig("222", "02", "1", "1", []string{"1"})
	if err != nil {
		t.Fatalf("Failed to create access mobility config: %v", err)
	}
```

polling/nf_configuration_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	regChan := make(chan []nfConfigApi.AccessAndMobility, 1)
+	updateCtxChan := make(chan []nfConfigApi.AccessAndMobility, 1)
+	go StartPollingService(ctx, "http://dummy", regChan, updateCtxChan)
+	time.Sleep(initialPollingInterval)
+
+	select {
+	case result := <-updateCtxChan:
+		if !reflect.DeepEqual(result, expectedConfig) {
+			t.Errorf("Expected %+v, got %+v", expectedConfig, result)
```
gab-arrobo, line 82: Make this consistent, in some cases it is uppercased and in other it is not across this file
For example:

```suggestion
			t.Errorf("expected %+v, got %+v", expectedConfig, result)
```

service/amf_server.go
```diff
@@ -122,10 +123,11 @@ func (s *Server) HandleMessage(srv sdcoreAmfServer.NgapService_HandleMessageServ
 	return nil
 }
 
-func StartGrpcServer(port int) {
+func StartGrpcServer(ctx context.Context, port int) {
 	endpt := fmt.Sprintf(":%d", port)
 	fmt.Println("listen - ", endpt)
```
gab-arrobo, line 128: I think we should change this to a "regular" log

```suggestion
	logger.GrpcLog.Infof("AMF gRPC server is starting on port %s", endpt)
```

service/init.go
```diff
@@ -375,14 +286,70 @@ func (amf *AMF) Start() {
...
+	if factory.AmfConfig.Configuration.Telemetry != nil && factory.AmfConfig.Configuration.Telemetry.Enabled {
+		var tp *sdktrace.TracerProvider
+		tp, err = tracing.InitTracer(ctx, tracing.TelemetryConfig{
+			OTLPEndpoint:   factory.AmfConfig.Configuration.Telemetry.OtlpEndpoint,
+			ServiceName:    "amf",
+			ServiceVersion: factory.AmfConfig.Info.Version,
+			Ratio:          *factory.AmfConfig.Configuration.Telemetry.Ratio,
+		})
+		if err != nil {
+			logger.InitLog.Panic("could not initialize tracer", zap.Error(err))
```
gab-arrobo, line 308: Why do you need to use `zap.Error`? Also, why not just using `Fatalf` instead of `Panic`?

service/init.go
```diff
@@ -375,14 +286,70 @@ func (amf *AMF) Start() {
...
+		}
+		logger.InitLog.Infoln("tracer initialized successfully")
+		defer func() {
+			err = tp.Shutdown(ctx)
+			if err != nil {
+				logger.InitLog.Error("failed to shutdown tracer", zap.Error(err))
```
gab-arrobo, line 314: Why do you need to use `zap.Error`?

```suggestion
				logger.InitLog.Errorf("failed to shutdown tracer", err)
```

service/init.go
```diff
@@ -375,14 +286,70 @@ func (amf *AMF) Start() {
...
+			case <-ctx.Done():
+				return
+			case cfg := <-contextUpdateChan:
+				factory.ConfigLock.Lock()
+				err = amfContext.UpdateAmfContext(self, cfg)
+				factory.ConfigLock.Unlock()
+				if err != nil {
+					logger.PollConfigLog.Errorf("AMF context update failed: %v", err)
+				} else {
+					logger.PollConfigLog.Debugf("AMF context updated from WebConsole config")
```
gab-arrobo, line 348:
```suggestion
					logger.PollConfigLog.Debugln("AMF context updated from WebConsole config")
```

service/init.go
```diff
@@ -466,75 +409,26 @@ func (amf *AMF) Start() {
...
-			logger.InitLog.Errorf("AMF start error: %+v", err)
-		}
-		wg.Done()
-	}()
-
-	wg.Wait()
-
-	return err
+	return nil
 }
```
gab-arrobo, line 413: Why having this function that the only thing that it does is `return nil`

polling/nf_configuration.go
```diff
@@ -0,0 +1,144 @@
...
+				case contextUpdateChannel <- newAccessMobilityConfig:
+				default:
+					logger.PollConfigLog.Warn("contextUpdateChan full, dropping config")
+				}
+				if plmnListIsChanged(poller.currentAccessAndMobilityConfig, newAccessMobilityConfig) {
+					logger.PollConfigLog.Debugf("Supported PLMN list changed")
+					select {
+					case registrationChannel <- newAccessMobilityConfig:
+					default:
+						logger.PollConfigLog.Warn("registrationChan full, dropping config")
```
gatici, line 71: Here we need to change the order of calls. When plmnListIsChanged(poller.currentAccessAndMobilityConfig, newAccessMobilityConfig) is called  poller.currentAccessAndMobilityConfig should include the old config. But line 60 (poller.currentAccessAndMobilityConfig = deepcopy.Copy(newAccessMobilityConfig).([]nfConfigApi.AccessAndMobility)) already replaces it with new config. We can write the flow like this:

```
			if !reflect.DeepEqual(newAccessMobilityConfig, poller.currentAccessAndMobilityConfig) {
				logger.PollConfigLog.Infof("Access and Mobility config changed. New Access and Mobility: %+v", newAccessMobilityConfig)
				// Check if PLMN list changed before updating the current config
				if plmnListIsChanged(poller.currentAccessAndMobilityConfig, newAccessMobilityConfig) {
					logger.PollConfigLog.Debugf("Supported PLMN list changed")
					select {
					case registrationChannel <- newAccessMobilityConfig:
					default:
						logger.PollConfigLog.Warn("registrationChan full, dropping config")
					}
				}
				// Now update poller state
				poller.currentAccessAndMobilityConfig = deepcopy.Copy(newAccessMobilityConfig).([]nfConfigApi.AccessAndMobility)
				select {
				case contextUpdateChannel <- newAccessMobilityConfig:
				default:
					logger.PollConfigLog.Warn("contextUpdateChan full, dropping config")
				}
			} else {
				logger.PollConfigLog.Debugf("Access and Mobility config did not change %+v", newAccessMobilityConfig)
			}
```

service/init.go
```diff
@@ -375,14 +286,70 @@ func (amf *AMF) Start() {
...
+			if err != nil {
+				logger.InitLog.Error("failed to shutdown tracer", zap.Error(err))
+			} else {
+				logger.InitLog.Infoln("tracer shutdown successfully")
+			}
+		}()
+	}
+
+	registrationChan := make(chan []nfConfigApi.AccessAndMobility, 1)
+	contextUpdateChan := make(chan []nfConfigApi.AccessAndMobility, 1)
```
gatici, line 322:
```suggestion
	registrationChan := make(chan []nfConfigApi.AccessAndMobility, 100)
	contextUpdateChan := make(chan []nfConfigApi.AccessAndMobility, 100)
```
We need to increase the channel buffer to a reasonable number as we use `select case` if the buffer is full, message will be lost.  Increasing channel buffer will help here to store the configs which are waiting for processing.

consumer/nf_management.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/omec-project/amf/logger"
+	"github.com/omec-project/amf/util"
+	"github.com/omec-project/openapi"
+	"github.com/omec-project/openapi/Nnrf_NFManagement"
+	"github.com/omec-project/openapi/models"
+	"github.com/omec-project/openapi/nfConfigApi"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+func getNfProfile(amfContext *amfContext.AMFContext, accessAndMobilityConfig []nfConfigApi.AccessAndMobility) (profile models.NfProfile, err error) {
```
patriciareinoso, line 25: I see that after polling, we update the `PlmnSupportList`, `SupportTaiLists` and 
`ServedGuamiList` in the AMF context.

the NRF registration is triggered only when there is a change in the PLMN list.

This function takes the PLMN list coming from the polling -> nrf registration execution flow, but the TAI and GUAMI list comes from the polling -> update amf context execution flow.

All the parameters should come from the execution flow and the registration to the NRF should be triggered whenever there is a change in any of the fields used for to build the profile

patriciareinoso (changes requested):

consumer/nf_management.go
dariofaccin: This comment made me think. Since the NF profile contains both supported PLMNs and supported TAIs (tuple of PLMN and TAC), it would make sense to trigger the registration whenever at least of them changes, not only the PLMNs. What do you think? Also @gatici

service/init.go
gruyaume: I have zero opinion on this, as far as I know those options are all ok fine.

factory/factory.go
```diff
@@ -12,128 +12,48 @@ package factory
...
+	"sync"
 
 	"github.com/omec-project/amf/logger"
 	"gopkg.in/yaml.v2"
 )
 
-var AmfConfig Config
+var (
+	AmfConfig  Config
+	ConfigLock sync.RWMutex
```
patriciareinoso, line 25: shouldn't this lock be in the context file?

consumer/nf_mangement.go
```diff
@@ -1,278 +0,0 @@
...
-	profile.NfInstanceId = context.NfId
-	profile.NfType = models.NfType_AMF
-	profile.NfStatus = models.NfStatus_REGISTERED
-	var plmns []models.PlmnId
-	for _, plmnItem := range context.PlmnSupportList {
-		plmns = append(plmns, plmnItem.PlmnId)
-	}
-	if len(plmns) > 0 {
-		profile.PlmnList = &plmns
-		// TODO: change to Per Plmn Support Snssai List
```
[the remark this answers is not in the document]
dariofaccin: Missed by mistake. I will also migrate to `PerPlmnSnssaiList` since we have the information.

polling/nf_configuration.go
```diff
@@ -0,0 +1,161 @@
...
+		case <-time.After(interval):
+			newAccessMobilityConfig, err := fetchAccessAndMobilityConfig(&poller, pollingEndpoint)
+			if err != nil {
+				interval = minDuration(interval*time.Duration(pollingBackoffFactor), pollingMaxBackoff)
+				logger.PollConfigLog.Errorf("Polling error. Retrying in %v: %+v", interval, err)
+				continue
+			}
+			interval = initialPollingInterval
+			if !reflect.DeepEqual(newAccessMobilityConfig, poller.currentAccessAndMobilityConfig) {
+				logger.PollConfigLog.Infof("Access and Mobility config changed. New Access and Mobility: %+v", newAccessMobilityConfig)
```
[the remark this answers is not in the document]
dariofaccin: I updated to set `Rcvd` parameter from the `UpdateAmfContext` func, where all the updates regarding configuration are done. Here there is also a lock mechanism in place.

polling/nf_configuration.go
```diff
@@ -0,0 +1,161 @@
...
+			newAccessMobilityConfig, err := fetchAccessAndMobilityConfig(&poller, pollingEndpoint)
+			if err != nil {
+				interval = minDuration(interval*time.Duration(pollingBackoffFactor), pollingMaxBackoff)
+				logger.PollConfigLog.Errorf("polling error. Retrying in %v: %+v", interval, err)
+				continue
+			}
+			interval = initialPollingInterval
+			if !reflect.DeepEqual(newAccessMobilityConfig, poller.currentAccessAndMobilityConfig) {
+				logger.PollConfigLog.Infof("Access and Mobility config changed. New Access and Mobility: %+v", newAccessMobilityConfig)
+				if plmnListOrTaiHasChanged(poller.currentAccessAndMobilityConfig, newAccessMobilityConfig) {
```
patriciareinoso, line 61: do we really need to check the changes at this granularity? 
each field in the `AccessAndMobility` struct is needed to build the NF profile, if any of them changes, the AMF re registers, so the `reflect.DeepEqual` should be enough?

service/init.go
```diff
@@ -557,254 +440,12 @@ func (amf *AMF) Terminate() {
...
-				// stop keepAliveTimer if its running and start the timer
-				amf.StartKeepAliveTimer(prof)
-				self.NfId = nfId
-				logger.CfgLog.Infoln("sent Register NF Instance with updated profile")
-			}
-		}
-	}
-}
-
 func UeConfigSliceDeleteHandler(ctx ctxt.Context, supi, sst, sd string, msg interface{}) {
```
patriciareinoso, line 447: all the functions below can be removed

gab-arrobo:
> Please run a go mod tidy we should see the proto dependency removed

Not really, because the `amf` runs a GRPC server for the `sctplb` to connect when it is deployed/used

context/context.go
```diff
@@ -633,3 +635,106 @@ func (context *AMFContext) Reset() {
...
+		factory.ConfigLock.Unlock()
+	}()
+	logger.ContextLog.Infoln("processing config update from polling service")
+	if len(newConfig) == 0 {
+		logger.ContextLog.Warnln("received empty access and mobility config, clearing dynamic AMF context")
+		amfContext.SupportTaiLists = amfContext.SupportTaiLists[:0]
+		amfContext.PlmnSupportList = amfContext.PlmnSupportList[:0]
+		amfContext.ServedGuamiList = amfContext.ServedGuamiList[:0]
+		factory.AmfConfig.Configuration.SliceTaiList = make(map[string][]models.Tai, 0)
+		factory.AmfConfig.Rcvd = false
```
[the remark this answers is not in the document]
gab-arrobo: Agree

context/context.go
```diff
@@ -633,3 +635,103 @@ func (context *AMFContext) Reset() {
...
+		}
+		sortSNssaiList(snssaiList)
+		plmnSupportItem := factory.PlmnSupportItem{
+			PlmnId:     plmn,
+			SNssaiList: snssaiList,
+		}
+		plmnSupportItemList = append(plmnSupportItemList, plmnSupportItem)
+		newGuami := models.Guami{
+			PlmnId: &plmn,
+			AmfId:  "cafe00",
```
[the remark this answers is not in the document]
dariofaccin: I was looking at [this](https://github.com/omec-project/amf/blob/268884ecb660b5f35d49eae4c43a163741fc94c5/service/init.go#L676).
@patriciareinoso maybe this should be part of the Access and Mobility configuration?

context/context.go
```diff
@@ -30,7 +32,7 @@ var (
 	tmsiGenerator                    *idgenerator.IDGenerator = nil
 	amfUeNGAPIDGenerator             *idgenerator.IDGenerator = nil
 	amfStatusSubscriptionIDGenerator *idgenerator.IDGenerator = nil
-	mutex                            sync.Mutex
```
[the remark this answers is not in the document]
gab-arrobo: Why does it have to be `RWMutex`? I am asking this because I see only two places/functions where `AmfContextMutex` (before: `mutex`) is being used: `NewAmfUe` and `UpdateAmfContext` and both of them seem to be only for writing. I do not see `RLock` being used. So, what is the benefit of changing it from `Mutex` to `RWMutex`?

context/context.go
```diff
@@ -32,7 +32,7 @@ var (
 	tmsiGenerator                    *idgenerator.IDGenerator = nil
 	amfUeNGAPIDGenerator             *idgenerator.IDGenerator = nil
 	amfStatusSubscriptionIDGenerator *idgenerator.IDGenerator = nil
-	AmfContextMutex                  sync.Mutex
+	AmfContextMutex                  sync.RWMutex
```
gab-arrobo, line 35: This is only used locally, so no need to exported, correct?
```suggestion
	amfContextMutex                  sync.RWMutex
```
Also, see my comment about using `Mutex` or `RWMutex`

context/context.go
```diff
@@ -112,7 +113,12 @@ type SecurityAlgorithm struct {
...
-func NewPlmnSupportItem() (item factory.PlmnSupportItem) {
+type PlmnSupportItem struct {
+	PlmnId     models.PlmnId   `yaml:"plmnId"`
+	SNssaiList []models.Snssai `yaml:"snssaiList,omitempty"`
+}
+
+func NewPlmnSupportItem() (item PlmnSupportItem) {
 	item.SNssaiList = make([]models.Snssai, 0, MaxNumOfSlice)
 	return
 }
```
gab-arrobo, line 124: I do not see this function being used anywhere in the codebase. Should we delete it?
```suggestion
```

factory/factory.go
gab-arrobo, line 25: Given that some of the parameters were moved to the `context`, ss this variable still needed?

service/init.go
```diff
@@ -141,91 +125,9 @@ func (amf *AMF) Initialize(ctx ctxt.Context, c *cli.Command) error {
...
-		logger.AppLog.Infoln("config file changed:", e.Name)
-		if err := factory.UpdateConfig(factory.AmfConfig.CfgLocation); err != nil {
-			logger.AppLog.Errorln("error in loading updated configuration")
-		} else {
-			self := context.AMF_Self()
-			util.InitAmfContext(self)
-			logger.AppLog.Infoln("successfully updated configuration")
-		}
-	})
-}
```
gab-arrobo, line 227: Given that this function is removed, I suspect that the use of `viper` becomes obsolete. So, should we remove the following code from the `Initialize` function above?
```
	if _, err := os.Stat(absPath); err == nil {
		viper.SetConfigFile(absPath)
		viper.SetConfigType("yaml")
		err = viper.ReadInConfig() // Find and read the config file
		if err != nil {            // Handle errors reading the config file
			return err
		}
	} else if os.IsNotExist(err) {
		logger.AppLog.Errorln("file %s does not exists", absPath)
		return err
	}
```

util/init_context.go
```diff
@@ -38,93 +38,93 @@ func InitDrsm() (drsm.DrsmInterface, error) {
...
-		context.Name = configuration.AmfName
+		amfContext.Name = configuration.AmfName
 	}
 	if configuration.NgapIpList != nil {
-		context.NgapIpList = configuration.NgapIpList
+		amfContext.NgapIpList = configuration.NgapIpList
 	} else {
-		context.NgapIpList = []string{"127.0.0.1"} // default localhost
+		amfContext.NgapIpList = []string{"127.0.0.1"} // default localhost
 	}
```
gab-arrobo, line 56:
```suggestion
	amfContext.NgapIpList = []string{"127.0.0.1"} // default localhost
	if configuration.NgapIpList != nil {
		amfContext.NgapIpList = configuration.NgapIpList
	}
```

gab-arrobo (commented):
Here is some additional feedback

gab-arrobo (commented):
<ignore-comment>

context/context.go
```diff
@@ -110,7 +113,12 @@ type SecurityAlgorithm struct {
 	CipheringOrder []uint8 // slice of security.AlgCipheringXXX
 }
 
-func NewPlmnSupportItem() (item factory.PlmnSupportItem) {
+type PlmnSupportItem struct {
+	PlmnId     models.PlmnId   `yaml:"plmnId"`
+	SNssaiList []models.Snssai `yaml:"snssaiList,omitempty"`
+}
```
gab-arrobo, line 119: BTW, any reason to define this struct and not using `PlmnSnssai` from `openapi` repo (https://github.com/omec-project/openapi/blob/main/models/model_plmn_snssai.go), which is pretty much the same?

context/context.go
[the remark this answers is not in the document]
patriciareinoso: Maybe, but since it was hardcoded in the code, it was not taken into account for the initial design.  Also, in the webconsole, we do not have a way to parameterized this parameter. To me it makes more sense to keep it in the AMF config file. But if we scale the AMF, should all of them use the same or a different ? there are many things to take into account to decide where to put it

patriciareinoso:
> > Please run a go mod tidy we should see the proto dependency removed
> 
> Not really, because the `amf` runs a GRPC server for the `sctplb` to connect when it is deployed/used. However, you do need to run `go mod tidy` to update other dependencies (one dependency moves from "direct" to "indirect")

I see, but the `	protos "github.com/omec-project/config5g/proto/sdcoreConfig"` which was used for the webconsole grpc, should dissappear, right?

context/context.go
patriciareinoso: Agree

dariofaccin: Updated to leverage `PlmnSnssai`. Beware that this struct allows for PLMNs to be `nil` (this should not happen in real scenarios).

context/context.go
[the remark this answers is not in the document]
gab-arrobo: I think the `amfid` can be provided through the initial yaml config file. What do you think @thakurajayL?

context/context.go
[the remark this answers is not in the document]
gab-arrobo: Should we add the reading locks? should they be part of this PR? or should we take care of that in a different PR?

go.mod
```diff
@@ -31,6 +28,11 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require (
+	github.com/fsnotify/fsnotify v1.9.0 // indirect
+	go.opencensus.io v0.24.0 // indirect
+)
+
 require (
```
gab-arrobo, line 36: I do not think we should have more than one "indirect" block of dependencies. I would suggest moving these 2 indirect imports into the same block as the other block of indirect imports and then run again `go mod tidy` to properly sort them
```suggestion
require (
	github.com/fsnotify/fsnotify v1.9.0 // indirect
	go.opencensus.io v0.24.0 // indirect
```

gab-arrobo (commented):
Overall, it looks good to me. We just need to finalize the discussion about the `amfId` and the discussion about the `RLock()` for the `RWMutex`

gab-arrobo:
> @gab-arrobo regarding the `amfId`, if we don't want to support different IDs for different PLMNs, I could add a field in the configuration yaml to allow configuring it. Then helm charts will be updated accordingly to set it if needed. Regarding the locking, I think a better approach would be to revert to regular `mutex` and implement the `RWLock` appropriately in a future PR. Let me know so I can finalize this.

Yes, please add the amfId as a configuration parameter (yaml) such that we can have e control of it during deployment. Regarding the mutex, let's keep the "regular" mutex and please open an issue such that we do not forget about the rwmutex and we address it in a future PR.

@thakurajayL, @ghislainbourgeois, @patriciareinoso, @gatici, I think we are almost done with this PR. Please provide feedback (if any) before we merge it. Thanks!

gatici (approved):

PR #473, opened 2025-06-25
feat: add tracing support with OpenTelemetry

gruyaume:
# Description

Here we add tracing support to the AMF, making it possible to trace both NGAP and NAS messages with OpenTelemetry traces. This feature is optional, disabled by default, and can be enabled via a new `telemetry` block in the configuration file.

This PR depends on the following PR's in libraries being merged first:
- https://github.com/omec-project/util/pull/167
- https://github.com/omec-project/openapi/pull/72
- https://github.com/omec-project/ngap/pull/33
- https://github.com/omec-project/nas/pull/112

## Usage

1. Stand up an OpenTelemetry compliant server
2. Start the AMF with the OpenTelemetry block described below

## Screenshots

Here we can see one of the trace visualization using Grafana Tempo

image

## Configuration file

We added the following block to the configuration file:

```yaml
configuration:
  ...
  telemetry:                                  # telemetry configuration
    enabled: true                             # Optional; defaults to false
    otlp_endpoint: "otel-collector.svc:4317"  # Mandatory if enabled=true
    ratio: 0.4                                # Optional; defaults to 1.0. If set to 0, AMF will assume 1.0. Use `enabled: false` to disable Telemetry.
```

## Future work

In the longer term the intent is to add distrubuted tracing to Aether SD-Core, making it possible to trace complete flows from the UE down to each network function. 

## Reference
- https://opentelemetry.io/
- https://grafana.com/oss/tempo/

PR #488, opened 2025-07-29
fix: properly shutdown tracer

dariofaccin:
This PR fixes the shutdown of the tracer introduced in #473.
When an interrupt or SIGTERM are provided, deferred functions are not called.
This PR moves the shutdown of the tracer in the `Terminate` function, so that it can be properly handled.


gab-arrobo (approved):
Overall, it looks good. Just one minor comment.

PR #489, opened 2025-08-02
Update/modify the codebase to be able to correctly handle Non-3GPP connections

gab-arrobo:


gmm/handler.go
```diff
@@ -1715,13 +1716,13 @@ func HandleServiceRequest(ctx ctxt.Context, ue *context.AmfUe, anType models.Acc
 			Procedure: context.OnGoingProcedureNothing,
 		})
 	} else if procedure != context.OnGoingProcedureNothing {
-		ue.GmmLog.Warnf("UE should not in OnGoing[%s]", procedure)
+		ue.GmmLog.Warnf("UE should not be in OnGoing[%s]", procedure)
```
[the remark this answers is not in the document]
gab-arrobo: It is already changed

thakurajayL:
It would be good to have PR description. May be its useful to see suggestion from co-pilot.

gmm/message/build.go
```diff
@@ -426,7 +429,16 @@ func BuildDeregistrationRequest(ue *context.RanUe, accessType uint8, reRegistrat
 			ProtocolDiscriminator: nasMessage.Epd5GSMobilityManagementMessage,
 			SecurityHeaderType:    nas.SecurityHeaderTypeIntegrityProtectedAndCiphered,
 		}
-		return nas_security.Encode(ue.AmfUe, m)
+		var anType models.AccessType
+		switch accessType {
+		case 0x01:
```
thakurajayL, line 434: 435 line shows models.AccessType and its assigned to anType. While case statement using 0x01, 0x02. Is it like we have not named variables correctly? Also is there any defined enum for 0x01, 0x02?

gab-arrobo: For some reason, I overlooked the `const` definition from the `nas` package. It is updated now. In a future PR, I am planning to update the `nas` module to define a type `AccessType` as follows such that it is easier to use/follow:
```diff
 // TS 24.501 9.11.3.11
+type AccessType uint8
+
 const (
-       AccessType3GPP    uint8 = 0x01
-       AccessTypeNon3GPP uint8 = 0x02
-       AccessTypeBoth    uint8 = 0x03
+       AccessType3GPP    AccessType = 0x01
+       AccessTypeNon3GPP AccessType = 0x02
+       AccessTypeBoth    AccessType = 0x03
 )
```

PR #517, opened 2025-09-10
Replace unmaintained dependency git.cs.nctu.edu.tw/calee/sctp with github.com/ishidawataru/sctp

gab-arrobo:


service/init.go
```diff
@@ -329,10 +329,7 @@ func (amf *AMF) Start() {
 
 	addr := fmt.Sprintf("%s:%d", self.BindingIPv4, self.SBIPort)
 
-	ngapHandler := ngap_service.NGAPHandler{
-		HandleMessage:      ngap.Dispatch,
-		HandleNotification: ngap.HandleSCTPNotification,
-	}
+	ngapHandler := ngap_service.NGAPHandler(ngap.Dispatch)
```
thakurajayL, line 332: Is it like new sctp module does not support notification? What happens in client (i.e. gNodeB) crashes/restarts

PR #661, opened 2026-04-16
fix(ngap): add nil guards for remaining 3 NGAP handlers

donivtech:
## Summary

Add post-loop nil checks for mandatory IEs in 3 NGAP handlers that lack
dispatcher-level protection via `FetchRanUeContext`:

- **HandleLocationReportingFailureIndication** — guard `cause`, `rANUENGAPID`
- **HandleRRCInactiveTransitionReport** — guard `rANUENGAPID`
- **HandleUplinkUEAssociatedNRPPATransport** — guard `rANUENGAPID`, `routingID`

Prevents nil pointer panics when a gNB omits mandatory IEs from these messages.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./ngap/...` — clean
- [x] `go test ./ngap/...` — passes

Signed-off-by: Vinod Patmanathan <vinod.patmanathan@forsway.com>

### DIFF
--- a/ngap/handler.go
+++ b/ngap/handler.go
@@ -1386,6 +1386,16 @@ func HandleLocationReportingFailureIndication(ran *context.AmfRan, message *ngap
 		}
 	}
 
+	if cause == nil {
+		ran.Log.Errorln("Cause is nil")
+		return
+	}
+
+	if rANUENGAPID == nil {
+		ran.Log.Errorln("RanUeNgapID is nil")
+		return
+	}
+
 	printAndGetCause(ran, cause)
 
 	ranUe = ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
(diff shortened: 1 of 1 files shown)


gab-arrobo:
@donivtech thanks for your contribution. Please rebase your PR

donivtech:
Thanks for the review @gab-arrobo and Copilot.

Pushed an updated version that addresses all feedback:

- **Rebased** onto latest `main`
- **Log message consistency**: aligned post-loop messages with the in-loop decode checks within each handler (`RanUeNgapID is nil`, `routingID is nil`, etc.)
- **Added `nRPPaPDU` nil guard**: good catch — it's marked `reject` criticality and the non-UE-associated variant already rejects missing NRPPaPDU, so the UE-associated variant should too.

gab-arrobo (approved):

PR #684, opened 2026-04-24
feat: add NR NTN access detection

donivtech:
Closes #683.

Detects NR Non-Terrestrial Network access by checking for the `NRNTN-TAI-Information` protocol extension (ID 287 per 3GPP TS 38.413) inside `UserLocationInformationNR.IEExtensions` received over NGAP, and tracks the result on the UE context.

### Changes
- New `context.NtnAccessInfo` struct (carries `Detected bool` today; shaped to hold future `satelliteBackhaulCat` / `geoSatelliteId` / NTN TAI list without reshaping `AmfUe`/`RanUe`).
- `AmfUe.NtnAccess`, `RanUe.NtnAccess`, `AmfUe.IsNtn()` helper.
- Detection inside `RanUe.UpdateLocation` (NR branch), running after the NR TAI/CGI parsing succeeds so a malformed message cannot leave NTN state out of sync with the stored location.
- Detection itself uses `ngap.HasUserLocationInformationNRExtension` from omec-project/ngap@v1.9.1 (added in [omec-project/ngap#87](https://github.com/omec-project/ngap/pull/87)).
- 6 table-driven cases in `context/ran_ue_test.go` covering the NTN-state semantics on `RanUe.updateNtnAccess`.

### Detection semantics — sticky
`Detected` is set true the first time the `NRNTN-TAI-Information` extension is observed for a UE during the current UE context. It is **not cleared** while the UE context lives. The extension is `PRESENCE optional` per 3GPP TS 38.413 and gNBs include it at their discretion — typically on context-establishing events (initial UE message, handover) but often not on routine uplink/location reports. Treating an absent extension as evidence of "not NTN" would cause `IsNtn()` to flap on routine NGAP traffic for a UE that genuinely is on NTN; sticky-on avoids that.

### Scope
Detection-only groundwork. No protocol behavior changes, no SBI payloads change, nothing reads the flag yet. Follow-up patches will extend `NtnAccessInfo` and wire NTN into the SBI path once `satelliteBackhaulCat` / `geoSatelliteId` are available in `omec-project/openapi`.

### Test plan
- [x] ` go build ./... ` clean
- [x] ` go vet ./... ` clean
- [x] ` go test ./... ` all pass (6 new sticky-semantics cases in `context/ran_ue_test.go`)
- [x] ` make lint ` (golangci-lint via Docker) — 0 issues

PR #685, opened 2026-04-24
fix: set RatType correctly during registration

donivtech:
### Problem
During a normal UE registration, ` ue.RatType ` is never set. The only code path that assigns it is ` CreateUEContextProcedure ` in ` producer/ue_context.go ` (inter-AMF relocation), and it does:

` ` ` go
ue.RatType = ueContextCreateData.UeContext.RestrictedRatList[0] // minItem = -1
` ` `

That is wrong on two counts:
- ` RestrictedRatList ` contains RAT types the UE is **not** allowed to use — it's the opposite of what ` RatType ` means.
- The list's minimum item count is -1 (i.e. it can be empty or nil), so indexing ` [0] ` can panic.

As a result, several SBI consumers (` Nsmf_PDUSession ` CreateSMContext, ` Namf_Communication ` UE context management, location reporting, MT notifications) send an empty ` ratType ` field.

### Fix
- Set ` ue.RatType ` in ` HandleRegistrationRequest ` based on access type:
  - 3GPP access + ` NrLocation ` → ` NR `
  - 3GPP access + ` EutraLocation ` → ` EUTRA `
  - Non-3GPP access → ` WLAN `
- Remove the buggy ` RestrictedRatList[0] ` assignment in ` CreateUEContextProcedure ` .

### Test plan
- [x] ` go build ./... ` clean
- [x] ` go vet ./... ` clean
- [x] ` go test ./gmm/... ./producer/... ` pass
- [x] ` make lint ` — 0 issues

### DIFF
--- a/gmm/handler.go
+++ b/gmm/handler.go
@@ -626,6 +626,18 @@ func HandleRegistrationRequest(ctx ctxt.Context, ue *context.AmfUe, anType model
 	ue.Location = ue.RanUe[anType].Location
 	ue.Tai = ue.RanUe[anType].Tai
 
+	switch anType {
+	case models.AccessType__3_GPP_ACCESS:
+		switch {
+		case ue.Location.NrLocation != nil:
+			ue.RatType = models.RatType_NR
+		case ue.Location.EutraLocation != nil:
+			ue.RatType = models.RatType_EUTRA
+		}
+	case models.AccessType_NON_3_GPP_ACCESS:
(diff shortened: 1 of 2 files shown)


donivtech:
Holding this until PR #684 (NR NTN access detection) gets maintainer feedback — naming/architecture decisions there can also apply here. Branch kept on the fork; will reopen when #684 lands.

PR #727, opened 2026-06-18
Handle high priority access service request as data

donivtech:
### What

Handle NAS Service Request service type 5 (high priority access,
TS 24.501 §9.11.3.50) as data instead of rejecting it.

Previously service type 5 fell through to the `default` case in
`HandleServiceRequest` and returned "service type[5] is not supported",
causing the AMF to send a Service Reject.

### Why

High priority access is a valid service type. The AMF implements no
MPS/priority semantics, so there is no priority behaviour to preserve;
handling it the same as service type 1 (data) is correct and avoids
rejecting otherwise-valid requests. This also accommodates modems that
send service type 5 for ordinary data sessions (e.g. Xiaomi M11 Lite 5G /
Qualcomm, some Quectel firmwares), which otherwise cannot recover from idle.

### How

Add a `case nasMessage.ServiceTypeHighPriorityAccess` that logs and falls
through into the existing `ServiceTypeData` handling. No config, no
behavioural change for any other service type.

### Testing

- `gofmt` clean, `golangci-lint` clean on the change

### DIFF
--- a/gmm/handler.go
+++ b/gmm/handler.go
@@ -2211,6 +2211,13 @@ func HandleServiceRequest(ctx ctxt.Context, ue *context.AmfUe, anType models.Acc
 				ue.ConfigurationUpdateMessage, &mobilityRestrictionList)
 			ue.ConfigurationUpdateMessage = nil
 		}
+	case nasMessage.ServiceTypeHighPriorityAccess:
+		// High priority access is a valid service type (TS 24.501 9.11.3.50).
+		// The AMF implements no MPS/priority semantics, so handle it as data
+		// rather than rejecting an otherwise-valid Service Request. Some
+		// modems also send service type 5 for ordinary data sessions.
+		ue.GmmLog.Infoln("handling Service Request with high priority access as data service")
+		fallthrough
 	case nasMessage.ServiceTypeData:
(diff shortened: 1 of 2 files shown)


gab-arrobo:
@donivtech, please address Copilot's comment and rebase PR. Thanks!

donivtech:
Thanks for the review. Addressed both points:

- Added a unit test `TestHandleServiceRequestHighPriorityAccessHandledAsData` in `gmm/handler_test.go`. It table-tests `HandleServiceRequest` for service type 5 (handled as data), service type 1 (baseline), and an emergency service type as a control (still rejected via the default case). It drives the UE down the data path's tracking-area-mismatch branch, which only the data case performs, so reaching it confirms service type 5 falls through to the data handling rather than returning "service type … is not supported".
- Rebased onto the latest `main`.

`go build/vet/test ./...`, `gofmt`, and `golangci-lint` are all clean.

donivtech:
Addressed the follow-up: the test now asserts the exact outcome per case instead of only checking for the absence of the "not supported" string — `err == nil` for the data and high-priority-access cases, and a non-nil error containing "is not supported" for the emergency control case. This removes the false-positive window where an unrelated error on the data path would still have passed. Tests, gofmt and golangci-lint remain clean.

gab-arrobo (approved):

PR #731, opened 2026-06-25
fix(ngap): add nil guards for missing mandatory NGAP-ID IEs in handlers

donivtech:
## Description

Several NGAP handlers dereference a mandatory NGAP-ID IE's `.Value` right after the IE-decoding loop. The in-loop checks only catch the "IE present but value is nil" case — an IE that is **entirely absent** from the `ProtocolIEs` list leaves the pointer `nil`, so a malformed/non-conformant peer can panic the AMF by omitting it.

The panic occurs in **two** places per handler: the lookup `RanUeFind…(id.Value)`, **and** the `No UE Context[…: %d]` error log immediately after it (which re-derefs the same ID and is reached precisely when the lookup returns nil).

This adds a post-loop `if <id> == nil { return }` guard before the first dereference in each affected handler.

> Note: simply swapping the lookup for the `findRanUeBy{Ran,Amf}NgapID` helper (as done in #732 for `FetchRanUeContext`) is **not** sufficient for the handlers — the helper returns `nil`, but the subsequent `No UE Context` log still derefs the ID and panics. The guard is needed because it returns *before* both the lookup and the log. (`FetchRanUeContext` is different: it returns the possibly-nil `ranUe` without logging the ID, so the helper is correct there.)

## Handlers guarded (10)

| Handler | ID |
|---|---|
| `HandleInitialUEMessage` (also `return` after the criticality-diagnostics `ErrorIndication`) | RANUENGAPID |
| `HandleInitialContextSetupResponse` | RANUENGAPID |
| `HandleInitialContextSetupFailure` | RANUENGAPID |
| `HandleUEContextReleaseRequest` (nil-safe dual AMFUENGAPID/RANUENGAPID lookup + drop nil-deref from its log) | both |
| `HandleUplinkNasTransport` | RANUENGAPID |
| `HandleUEContextReleaseComplete` | AMFUENGAPID |
| `HandlePDUSessionResourceReleaseResponse` | RANUENGAPID |
| `HandleHandoverCancel` | RANUENGAPID |
| `HandleNasNonDeliveryIndication` | RANUENGAPID |
| `HandleUERadioCapabilityInfoIndication` | RANUENGAPID |

The remaining handlers already guard their mandatory IDs (post-loop `== nil` guard, inline `if id != nil` wrapper, or the criticality-diagnostics `if len > 0 { return }` pattern), so they're left unchanged.

## Testing

- `golangci-lint run --new-from-rev=main` → `0 issues.`
- `gofmt` clean

### DIFF
--- a/ngap/handler.go
+++ b/ngap/handler.go
@@ -802,7 +802,7 @@ func HandleUplinkNasTransport(ctx ctxt.Context, ran *context.AmfRan, message *ng
 		return
 	}
 
-	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
 	if ranUe == nil {
 		ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
 		return
@@ -1051,7 +1051,7 @@ func HandleUEContextReleaseComplete(ctx ctxt.Context, ran *context.AmfRan, messa
 		return
 	}
 
-	ranUe := context.AMF_Self().RanUeFindByAmfUeNgapID(aMFUENGAPID.Value)
+	ranUe := findRanUeByAmfNgapID(ran, aMFUENGAPID)
 	if ranUe == nil {
 		ran.Log.Errorf("No RanUe Context[AmfUeNgapID: %d]", aMFUENGAPID.Value)
 		cause := ngapType.Cause{
@@ -1342,7 +1342,7 @@ func HandlePDUSessionResourceReleaseResponse(ctx ctxt.Context, ran *context.AmfR
 		return
 	}
 
-	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
 	if ranUe == nil {
 		ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
 		return
@@ -2526,7 +2526,7 @@ func HandleInitialContextSetupResponse(ctx ctxt.Context, ran *context.AmfRan, me
 		return
 	}
 
-	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
 	if ranUe == nil {
 		ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
 		return
@@ -2690,7 +2690,7 @@ func HandleInitialContextSetupFailure(ctx ctxt.Context, ran *context.AmfRan, mes
 		return
 	}
 
-	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
 	if ranUe == nil {
 		ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
 		return
@@ -4038,7 +4038,7 @@ func HandleHandoverCancel(ctx ctxt.Context, ran *context.AmfRan, message *ngapTy
 		return
 	}
 
-	sourceUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+	sourceUe := findRanUeByRanNgapID(ran, rANUENGAPID)
 	if sourceUe == nil {
 		ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
 		cause := ngapType.Cause{
@@ -4237,7 +4237,7 @@ func HandleNasNonDeliveryIndication(ctx ctxt.Context, ran *context.AmfRan, messa
 		return
 	}
 
-	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
 	if ranUe == nil {
 		ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
 		return
@@ -4799,7 +4799,7 @@ func HandleUERadioCapabilityInfoIndication(ran *context.AmfRan, message *ngapTyp
 		return
 	}
 
-	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
 	if ranUe == nil {
 		ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
 		return
--- a/ngap/handler.go
+++ b/ngap/handler.go
@@ -797,6 +797,10 @@ func HandleUplinkNasTransport(ctx ctxt.Context, ran *context.AmfRan, message *ng
 		ran.Log.Errorln("AMFUENGAPID IE missing from UplinkNasTransport")
 		return
 	}
+	if nASPDU == nil {
+		ran.Log.Errorln("NASPDU IE missing from UplinkNasTransport")
+		return
+	}
 
 	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
 	if ranUe == nil {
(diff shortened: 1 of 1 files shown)


ngap/handler.go
```diff
@@ -2744,12 +2761,15 @@ func HandleUEContextReleaseRequest(ctx ctxt.Context, ran *context.AmfRan, messag
...
+	var ranUe *context.RanUe
+	if aMFUENGAPID != nil {
+		ranUe = context.AMF_Self().RanUeFindByAmfUeNgapID(aMFUENGAPID.Value)
+	}
+	if ranUe == nil && rANUENGAPID != nil {
 		ranUe = ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
 	}
 	if ranUe == nil {
-		ran.Log.Errorf("No RanUe Context[AmfUeNgapID: %d]", aMFUENGAPID.Value)
+		ran.Log.Errorln("No RanUe Context for UEContextReleaseRequest")
```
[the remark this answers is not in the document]
gab-arrobo: I think this comment still applies. correct?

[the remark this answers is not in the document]
donivtech: Yes, resolved in 56e9d3f

donivtech:
Good catch. Fixed in c689af4. Added a post-loop nASPDU == nil guard in HandleUplinkNasTransport, HandleInitialUEMessage, and HandleNasNonDeliveryIndication. I also swept all the handlers touched in this PR for any other post-loop IE dereference and found none remaining.

ngap/handler.go
```diff
@@ -789,6 +789,19 @@ func HandleUplinkNasTransport(ctx ctxt.Context, ran *context.AmfRan, message *ng
...
+	if aMFUENGAPID == nil {
+		ran.Log.Errorln("AMFUENGAPID IE missing from UplinkNasTransport")
+		return
+	}
+	if nASPDU == nil {
+		ran.Log.Errorln("NASPDU IE missing from UplinkNasTransport")
+		return
+	}
+
 	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
```
gab-arrobo: We already have a helper function `findRanUeByRanNgapID` that checks whether rANUENGAPID is nil, why not using it?

```suggestion
	if aMFUENGAPID == nil {
		ran.Log.Errorln("AMFUENGAPID IE missing from UplinkNasTransport")
		return
	}
	if nASPDU == nil {
		ran.Log.Errorln("NASPDU IE missing from UplinkNasTransport")
		return
	}

	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
```

ngap/handler.go
```diff
@@ -1033,6 +1046,11 @@ func HandleUEContextReleaseComplete(ctx ctxt.Context, ran *context.AmfRan, messa
 		}
 	}
 
+	if aMFUENGAPID == nil {
+		ran.Log.Errorln("AMFUENGAPID IE missing from UEContextReleaseComplete")
+		return
+	}
+
 	ranUe := context.AMF_Self().RanUeFindByAmfUeNgapID(aMFUENGAPID.Value)
```
gab-arrobo:
```suggestion
	ranUe := findRanUeByAmfNgapID(ran, aMFUENGAPID)
```

ngap/handler.go
```diff
@@ -1319,6 +1337,11 @@ func HandlePDUSessionResourceReleaseResponse(ctx ctxt.Context, ran *context.AmfR
 		}
 	}
 
+	if rANUENGAPID == nil {
+		ran.Log.Errorln("RANUENGAPID IE missing from PDUSessionResourceReleaseResponse")
+		return
+	}
+
 	ranUe := ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
```
gab-arrobo: Should not we reuse the helper functions (`findRanUeByRanNgapID`, `findRanUeByAmfNgapID`) already defined? Or an alternative, what about expanding the arguments in these helper functions to include the additional logging details you are adding such as "from xxxxxxx"
```suggestion
	ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
```

gab-arrobo (commented):
@donivtech, thanks for your contribution. Please see a few comments below but those comments are applicable across the proposed changes. Let me know what you think. Thanks!

donivtech:
Switched these to the helpers as suggested (44fe599): findRanUeByRanNgapID / findRanUeByAmfNgapID for the lookups in UplinkNasTransport, UEContextReleaseComplete, PDUSessionResourceReleaseResponse, InitialContextSetupResponse/Failure, HandoverCancel, NasNonDeliveryIndication, and UERadioCapabilityInfoIndication.

In each of these the No UE Context[…: %d] log right after the lookup dereferences the same ID, and with the helper that line is now reached when the ID is nil — so I wrapped it in if id != nil to stay panic-safe (the helper already logs the nil-ID case).

Happy to instead expand the helpers to take a caller label (your "from X" idea) if you'd prefer that over the generic nil-ID log - just say

gab-arrobo:
> Switched these to the helpers as suggested ([44fe599](https://github.com/omec-project/amf/commit/44fe599009ca0683b520dde86503c586f1e0fa24)): findRanUeByRanNgapID / findRanUeByAmfNgapID for the lookups in UplinkNasTransport, UEContextReleaseComplete, PDUSessionResourceReleaseResponse, InitialContextSetupResponse/Failure, HandoverCancel, NasNonDeliveryIndication, and UERadioCapabilityInfoIndication.
> 
> In each of these the No UE Context[…: %d] log right after the lookup dereferences the same ID, and with the helper that line is now reached when the ID is nil — so I wrapped it in if id != nil to stay panic-safe (the helper already logs the nil-ID case).
> 
> Happy to instead expand the helpers to take a caller label (your "from X" idea) if you'd prefer that over the generic nil-ID log - just say

Hi @donivtech,
Sorry, I got confused why you needed to check `rANUENGAPID == nil`. based on your latest changes, I see that you still need to check that because of the log inside `if ranUe == nil`.
```go
	if rANUENGAPID == nil {
		ran.Log.Errorln("RANUENGAPID IE missing from UplinkNasTransport")
		return
	}
```

I think a better approach would be to keep the following check(s) and remove the if statements from inside `if ranUe == nil`. What do you think?
```go
	if rANUENGAPID == nil {
		ran.Log.Errorln("RANUENGAPID IE missing from UplinkNasTransport")
		return
	}
```

That is, something like this in all places where `rANUENGAPID.Value` is used inside a `if ranUe == nil`

```diff
diff --git a/ngap/handler.go b/ngap/handler.go
index ed83f62..4b7599e 100644
--- a/ngap/handler.go
+++ b/ngap/handler.go
@@ -789,6 +789,11 @@ func HandleUplinkNasTransport(ctx ctxt.Context, ran *context.AmfRan, message *ng
                }
        }

+       if rANUENGAPID == nil {
+               ran.Log.Errorln("RANUENGAPID IE missing from UplinkNasTransport")
+               return
+       }
+
        if aMFUENGAPID == nil {
                ran.Log.Errorln("AMFUENGAPID IE missing from UplinkNasTransport")
                return
@@ -800,9 +805,7 @@ func HandleUplinkNasTransport(ctx ctxt.Context, ran *context.AmfRan, message *ng

        ranUe := findRanUeByRanNgapID(ran, rANUENGAPID)
        if ranUe == nil {
-               if rANUENGAPID != nil {
-                       ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
-               }
+               ran.Log.Errorf("No UE Context[RanUeNgapID: %d]", rANUENGAPID.Value)
                return
        }

```


Thanks!

donivtech:
Yes, that is better— applied your suggestion in the latest push. Kept the explicit if id == nil { … return } guard before each lookup and removed the if id != nil wrapper from inside if ranUe == nil, while keeping the findRanUeBy*NgapID helper for the lookup itself. So it reads guard → helper → clean No UE Context[…: %d] log (the guard guarantees the ID is non-nil by that point). Also merged latest main in.

Applied across all eight: UplinkNasTransport, UEContextReleaseComplete, PDUSessionResourceReleaseResponse, InitialContextSetupResponse/Failure, HandoverCancel, NasNonDeliveryIndication, UERadioCapabilityInfoIndication.

ngap/handler.go
```diff
@@ -2744,12 +2788,24 @@ func HandleUEContextReleaseRequest(ctx ctxt.Context, ran *context.AmfRan, messag
 		}
 	}
 
-	ranUe := context.AMF_Self().RanUeFindByAmfUeNgapID(aMFUENGAPID.Value)
-	if ranUe == nil {
+	var ranUe *context.RanUe
+	if aMFUENGAPID != nil {
+		ranUe = context.AMF_Self().RanUeFindByAmfUeNgapID(aMFUENGAPID.Value)
+	}
```
gab-arrobo: Why not using `findRanUeByAmfNgapID`?
```suggestion
	ranUe := findRanUeByAmfNgapID(ran, aMFUENGAPID)
```

donivtech:
Good point , done. HandleUEContextReleaseRequest now uses findRanUeByAmfNgapID(ran, aMFUENGAPID) for the primary lookup (and findRanUeByRanNgapID(ran, rANUENGAPID) for the RAN-ID fallback), dropping the != nil wrappers since the helpers nil-check internally.

gab-arrobo (approved):

PR #732, opened 2026-06-25
fix(ngap): route remaining FetchRanUeContext lookups through nil-safe helper

donivtech:
## Description

#666 introduced the nil-safe helper `findRanUeByRanNgapID()` — it returns `nil` when the `RANUENGAPID` IE pointer is `nil` instead of dereferencing `.Value` — and converted most procedure-code cases in `FetchRanUeContext` to use it. Three cases were missed and still call `ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)` directly:

- `HandoverNotification`
- `HandoverPreparation`
- `PDUSessionResourceModify` (Response)

In each, the in-loop check only catches a `RANUENGAPID` IE that is **present with a nil value**. If the IE is **entirely absent** from the message, `rANUENGAPID` stays `nil` and the subsequent `.Value` dereference panics the AMF — the same malformed-input crash class #666 set out to close.

This routes those three lookups through `findRanUeByRanNgapID()` like the rest of the function, making `FetchRanUeContext` uniformly nil-safe.

## Testing

- `golangci-lint run --new-from-rev=main` → `0 issues.`
- `gofmt` clean

### DIFF
--- a/ngap/handler.go
+++ b/ngap/handler.go
@@ -247,7 +247,7 @@ func FetchRanUeContext(ran *context.AmfRan, message *ngapType.NGAPPDU) (*context
 					aMFUENGAPID = ie.Value.AMFUENGAPID
 				}
 			}
-			ranUe = ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+			ranUe = findRanUeByRanNgapID(ran, rANUENGAPID)
 		case ngapType.ProcedureCodeHandoverPreparation:
 			ngapMsg := initiatingMessage.Value.HandoverPreparation
 			for i := 0; i < len(ngapMsg.ProtocolIEs.List); i++ {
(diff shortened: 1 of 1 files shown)


gab-arrobo (approved):

gab-arrobo:
Should we make this change across the codebase? There are other 20+ occurrences of this
```diff
-			ranUe = ran.RanUeFindByRanUeNgapID(rANUENGAPID.Value)
+			ranUe = findRanUeByRanNgapID(ran, rANUENGAPID)
```

donivtech:
Agreed. It's worth doing across the codebase: beyond the 4 handlers in #731, there are 6 more that still panic on an absent ID IE (UplinkNasTransport, PDUSessionResourceReleaseResponse, HandoverCancel, NasNonDeliveryIndication, UERadioCapabilityInfoIndication, UEContextReleaseComplete). I'll do a sweep converting all the raw RanUeFindBy{Ran,AmfUe}NgapID(...Value) calls to the two helpers.

donivtech:
Implemented in #731. Expanded from the original 4 handlers to all 10 that actually panic on a missing NGAP-ID IE. Full breakdown's in the PR description.

PR #739, opened 2026-07-03
refactor: route AmfUe identity reads through the Get* accessors (follow-up to #735)

midwell:
## Summary

Follow-up to #735 (@gab-arrobo's request there): route the **remaining direct reads** of the `AmfUe` identity fields — `Supi`, `Pei`, `Gpsi`, `Guti`, `Tmsi`, `RegistrationType5GS` — through the `Get*` accessors that #735 introduced, so all identity access outside the `context` package goes through the `identityMu`-synchronized API.

## Dependency — stacked on #735

This **depends on #735**, which adds the `Get*`/`Set*`/`IdentitySnapshot` accessors consumed here. Please merge #735 first; until then its commits appear in this PR's diff and will drop out once #735 lands (I'll rebase if needed). #735 already converted every identity **write** (→ `Set*`) and every **cross-goroutine read** (→ `Get*`/`IdentitySnapshot`); this PR finishes the remaining same-goroutine reads.

## Scope

69 reads converted across `consumer/` (`am_policy`, `communication`, `sm_context`, `subscriber_data_management`, `ue_context_management`) and `gmm/` (`handler`, `message/build`, `sm`).

**No behavioural change.** These reads execute on the UE's own NAS-processing goroutine, serialized with the writers, so they were already race-free — routing them through `Get*` makes the access pattern uniform and stops a future cross-goroutine reader from silently reintroducing a race.

**Left as direct access — the `context` package itself.** It owns `AmfUe` and defines the accessors, so internal direct field access is idiomatic, and #735 already routed its cross-goroutine reads through `Get*`. (Standard encapsulation: accessors are for external callers.) Happy to convert those too if you'd prefer.

## Verification

`go build ./...` + `go vet ./...` clean and `golangci-lint v2.11.3` (`.golangci.yml`) → **0 issues**, in the `linux/amd64` container. The change is purely `field` → `GetField()`; the compiler confirms every converted receiver is an `*AmfUe`.

### DIFF
--- a/consumer/am_policy.go
+++ b/consumer/am_policy.go
@@ -26,7 +26,7 @@ func AMPolicyControlCreate(ctx context.Context, ue *amf_context.AmfUe, anType mo
 		attribute.String("http.method", "POST"),
 		attribute.String("nf.target", "pcf"),
 		attribute.String("net.peer.name", ue.PcfUri),
-		attribute.String("ue.supi", ue.Supi),
+		attribute.String("ue.supi", ue.GetSupi()),
 		attribute.String("ue.plmn.id", ue.PlmnId.Mcc+ue.PlmnId.Mnc),
 	)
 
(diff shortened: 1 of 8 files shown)


gab-arrobo:
@midwell,
Please resolve conflict and rebase PR. Thanks!

midwell:
> @midwell, Please resolve conflict and rebase PR. Thanks!

Rebased onto latest main. Since #735 is now merged, this PR is down to the single accessor-adoption commit (the earlier accessor-foundation commits dropped out). Also rebased #738 onto main while I was at it. Thanks!

gab-arrobo (approved):

PR #793, opened 2026-08-25
context, gmm: keep a pending N1N2 message readable, and stop storing from destroying it

midwell:
### What this fixes

Three defects on one path. Each of them alone leaves a UE that cannot be reached with downlink
traffic, and they compose — fixing any one still ends in a UE with no user plane, arrived at from
a different direction.

**1. The payload was already consumed before the message was stored.**

`N1N2Message` held only the request, whose binary parts are readers. The transfer procedure drains
them to decide what to do, so by the time the message was stored for a later page, the payloads
were gone. A paged UE then came back and received a `PDUSessionResourceSetupRequest` carrying an
**empty transfer** — connected, with no user plane, and nothing in the log to say why.

**2. Storing a context destroyed the message inside it.**

`MarshalJSON` materialised empty optional containers on the live `JsonData` — the same pointer the
pending message holds. So persisting a context emptied the pending message it was persisting.

**3. A stored context could not be read back at all.**

Those empty containers are written as `n1MessageClass:""` and `n2InformationClass:""`, and the
strict 3GPP enum decoders refuse an empty value — refusing the whole document with it.

This turned out to be a *class* of defect rather than two fields: after fixing both containers, a
round trip still failed on `ngKsi.tsc`, which is empty on any UE stored before authentication
finished. A test is what found that; inspection did not.

### What changes

- The payload bytes are captured alongside the request, and a `Payloads` accessor returns them,
  reading the request only as a fallback for a message stored before this change. A container
  declared with nothing behind it is now an error rather than an empty item on the wire.
- Storing serialises from a **copy**, and omits a container carrying no class rather than writing
  it empty.
- The read path keeps its strict decode and, **only when that fails**, drops empty values and
  retries once, logging that it did. This is safe for a reason worth stating: decoding treats an
  absent key and an empty string identically — the field is left at its zero value — and only the
  strict enum decoders distinguish them, by refusing everything. The write path stops producing
  such records, so the tolerance exists purely for what is already in deployed databases.
- `DbFetch` mapped *any* unmarshal error to nil, which the caller reports as "no document found".
  That is what hid this for so long: "no document found" describes a missing subscriber, not a
  record that is present and unreadable. They are now different log lines.
- Guards `MarshalJSON`'s unchecked `RanUe.Ran` dereference, which panicked when storing a context
  whose `RanUe` had no RAN — one restored from the database, or one not yet attached.

### Testing

`linux/amd64`, in the images CI pins:

- `go test -race ./...` — **all packages green** in `golang:1.25.0`
- `golangci-lint run` — **0 issues** in `golangci/golangci-lint:v2.11.3`

The tests are mutation-checked rather than assumed: restoring the pre-fix container
materialisation makes `TestStoredContextWithPendingMessageRoundTrips` fail with
`Payloads on a restored message: N1MessageContainer present but BinaryDataN1Message is missing`,
and `TestSerialisablePendingMessageOmitsContainersWithNoClass` fail as well. Both pass with the
fix.

Verified end to end on a live deployment: a page now produces an `InitialContextSetupRequest`
whose PDU session transfer carries 4 IEs and 36 bytes, where it previously carried none, and the
UE's user plane returns 18 ms after the page.

### Note for reviewers

The read-path tolerance is the part I would most like a second opinion on. It is deliberately
scoped to the persistence path and to a retry *after* a strict decode has already failed, so a
well-formed record never goes near it. The alternative — relaxing the shared enum decoders in
`openapi/models` to accept `""` — would weaken wire validation for every NF, where an empty class
really is invalid. If you would rather this healed records by a migration instead, I am happy to
do it that way; the writer fix stands either way.

### DIFF
--- a/context/amf_ue.go
+++ b/context/amf_ue.go
@@ -12,7 +12,9 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/hex"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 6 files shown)


gab-arrobo:
@midwell, please address conflict and rebase PR. Thanks!

gab-arrobo (approved):

PR #794, opened 2026-08-25
ngap, producer: page a UE instead of sending N2 into a connection that is gone

midwell:
> **Stacked on #793.** The diff shown here includes that PR's commit until it merges; the commit
> to review is the second one. They are separable but both touch `context/amf_ue.go`, so this one
> is based on it to avoid a conflict.

### What this fixes

An idle UE stayed unreachable in two situations that end the same way: the AMF decides to send an
N2 message rather than page, the message does not arrive, and the SMF is told
`N1N2_TRANSFER_INITIATED`. Nothing waits and nothing retries, so the UE becomes reachable again
only when it happens to come back on its own.

**1. A record of a past connection is not a connection.**

A context restored from the database carries a `RanUe`, and that `RanUe` carries an `AmfRan` — but
only the fields that survive serialisation. `Conn`, `Amf2RanMsgChan` and the logger are `json:"-"`
and come back nil. Treating that as connected sent N2 into an association that no longer existed:

```
send PDU Session Resource Setup Request {amf_ue_ngap_id: AMF_UE_NGAP_ID:0}
send error, gNB may have been lost: nil pointer dereference
```
…and **zero Paging frames**, with HTTP 200 returned to the SMF.

`CmConnect` cannot be the test here. It answers a wider question — "is this UE associated over
this access type" — and is used for lookups where a stale `RanUe` is perfectly fine; `GetAnType`
is one such caller. Making it strict turns a Namf_Location 200 into a 404, which an existing test
catches. So this adds `HasLiveRanConnection` and uses it only at the three sites that choose
between an N2 send and a page.

Two things worth flagging, because both would have shipped as a confident non-fix:

- The obvious guard, `Ran == nil`, is **false** for a restored context — it is not empty, only its
  transport is missing.
- Under **SCTP load balancing** the AMF holds no socket at all; messages leave through a channel.
  A liveness test written against `Conn` alone would report every UE idle and break those
  deployments while fixing ours.

**2. `SendToRan` could fail in ways the caller could not see.**

Its own nil guards logged through the `AmfRan`'s logger, which a restored context does not have —
so the guard panicked and the `recover` attributed an internal fault to the gNB. And under SCTP
load balancing the send is a channel send: a restored context's channel is nil, so the caller did
not fail, it **stopped there for ever**, holding the message it meant to page for. The `recover`
stays — a dead peer should not take the AMF down — but it no longer launders an internal fault
into "gNB may have been lost".

**3. The gNB can refuse a message that was sent successfully.**

`HandleErrorIndication` decoded, logged, and returned on a `// TODO`. So a gNB answering
`unknown-local-UE-NGAP-ID` produced nothing at all: the write had succeeded and the procedure had
already answered the SMF. Observed as a `PDUSessionResourceSetupRequest`, that indication, and the
UE recovering 35 s later by its own service request.

For the two causes that mean the peer does not recognise the id, the stale context is now released
and the pending message is paged for. Releasing it **first** is what makes the paging decision
correct — the UE has to stop looking connected to a gNB that just said it does not know it, or the
next transfer repeats the failure identically.

### Testing

`linux/amd64`, in the images CI pins:

- `go test -race ./...` — **all packages green** in `golang:1.25.0`
- `golangci-lint run` — **0 issues** in `golangci/golangci-lint:v2.11.3`

Mutation-checked rather than assumed. Removing the nil-channel guard hangs
`TestSendToRanReturnsForARestoredRan` for its full timeout; removing the ownership guard fails
`TestErrorIndicationFromAnotherGnbIsIgnored`; and removing the *call site* from
`HandleErrorIndication` fails `TestHandleErrorIndicationRecoversThroughTheRealMessage` — that last
test exists because an earlier version asserted only on the helper, which passes just as happily
when nothing calls it, the same shape as the defect itself.

On a live deployment the scenario that produced `AMF_UE_NGAP_ID:0` and zero Paging now produces
`send Paging to TAI({Mcc:208 Mnc:93}, Tac:000001)` and two Paging frames carrying the restored
UE's 5G-S-TMSI. Five further paging cycles all completed, page to restored user plane in 18 ms.

### Note for reviewers

- **The `ErrorIndication` recovery is scoped deliberately.** It acts only on
  `unknown-local-UE-NGAP-ID` / `inconsistent-remote-UE-NGAP-ID`, and only on a `RanUe` the sending
  gNB actually holds — the AMF-assigned id is unique across the AMF, not within one gNB, so an
  unrelated peer must not be able to have a subscriber's context released on its word.
- **No lock would have fixed #3.** The AMF's view was self-consistent; the peer had released the
  UE. Repairing the outcome is the remedy rather than trying to win a race that is not the AMF's
  to win. I mention it because "make the decision and the send atomic" is the natural first
  suggestion and it does not help here.
- **One half deliberately not built:** a locally failed write still reports success to the SMF.
  A failed write means the association is gone, which tears down the `AmfRan` and its `RanUe`s, so
  the next transfer pages correctly — and threading an error back through the 29 call sites of
  `SendToRan`/`SendToRanUe` to recover one in-flight message did not look like a good trade. Happy
  to do it if you disagree.

### DIFF
--- a/context/amf_ue.go
+++ b/context/amf_ue.go
@@ -668,6 +668,40 @@ func (ue *AmfUe) CmConnect(anType models.AccessType) bool {
 	return true
 }
 
+// HasLiveRanConnection reports whether an N2 message for this UE can actually reach a
+// RAN node right now.
+//
+// This is a narrower question than CmConnect, which answers "is this UE associated
+// over this access type" and is used for lookups where a stale RanUe is acceptable --
+// GetAnType is one such caller.
+//
+// A record of a past connection is not a connection. A context restored from the
+// database carries a RanUe, and that RanUe carries an AmfRan, but only the fields that
+// survive serialisation: Conn and Amf2RanMsgChan are json:"-" and come back nil, as
+// does the logger. Treating that as connected sent N2 messages into an association
+// that no longer existed, panicked in the send path on the nil logger, had the panic
+// swallowed by a recover that blamed the gNB, and still answered the SMF as though the
+// message had gone -- so an idle UE stayed unreachable after every AMF restart,
+// because nothing ever paged it.
+func (ue *AmfUe) HasLiveRanConnection(anType models.AccessType) bool {
+	ue.Mutex.Lock()
+	defer ue.Mutex.Unlock()
+
+	ranUe, ok := ue.RanUe[anType]
+	if !ok || ranUe == nil || ranUe.Ran == nil {
+		return false
+	}
+
+	// Under SCTP load balancing the AMF holds no socket of its own: messages leave
+	// through a channel, so Conn being nil means nothing there and the channel is what
+	// has to exist.
+	if AMF_Self().EnableSctpLb {
+		return ranUe.Ran.Amf2RanMsgChan != nil
+	}
+
+	return ranUe.Ran.Conn != nil
+}
+
 func (ue *AmfUe) CmIdle(anType models.AccessType) bool {
 	return !ue.CmConnect(anType)
 }
--- a/context/cm_state_test.go
+++ b/context/cm_state_test.go
@@ -1,0 +1,117 @@
+// SPDX-FileCopyrightText: 2026 Forsway Scandinavia AB
+// SPDX-License-Identifier: Apache-2.0
+
+package context
+
+import (
+	"net"
+	"testing"
+
+	"github.com/omec-project/amf/protos/sdcoreAmfServer"
+	"github.com/omec-project/openapi/v2/models"
+)
+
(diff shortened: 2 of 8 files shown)


gab-arrobo:
@midwell, please address conflict and rebase PR. Thanks!

midwell:
Rebased — `1bc71ca`. The conflict is gone: #793 merged, so its commit dropped out of this branch and what remains is one commit on `main`, eight files, entirely this change.

Rechecked on the current base: build, the `ngap`, `producer`, `context` and `gmm` tests, and `golangci-lint v2.13.2` all clean.

gab-arrobo (approved):

PR #795, opened 2026-08-25
feat: signal the extended NAS SM timer indication, from the orbit rather than from being satellite

midwell:
## What

The AMF tells the SMF when a UE's access warrants the extended NAS session management timer values,
by setting `ExtendedNasSmTimerInd` in `SmContextCreateData`.

The field exists on the model and nothing has ever set it. 99 lines, of which 66 are tests.

## Why

TS 24.501 table 10.3.2, *Timers of 5GS session management – SMF side*, gives the NAS session
management timers a base value and a satellite one — T3591 is 16 s or 22 s — and the SMF has to know
which applies to a given session. Subclause 4.23.4 is what routes it there: *"If the use of extended
NAS timer for access via a satellite NG-RAN cell is indicated by the AMF … the SMF shall calculate
the value of the applicable NAS timer indicated in table 10.3.2 for access via a satellite NG-RAN
cell."* The specified signal is this indication from the AMF. With nothing setting it, an SMF that reads it can
only fall back to configuration, which cannot distinguish a satellite cell from a terrestrial one on
the same AMF.

The consumer is [omec-project/smf#623](https://github.com/omec-project/smf/pull/623), which reads it
when resolving T3591.

## The narrower predicate is the point

`AmfUe.IsNtn()` already exists and answers "is this access non-terrestrial". Reusing it would be
wrong, and that is the whole substance of this change.

**NOTE 5 of table 10.3.2** names **NR(MEO)** and **NR(GEO)**, and no other access — table 10.3.1
says the same for the UE side in its NOTE 7. (Corrected: an earlier revision of this description and
of the code comments attributed that note to subclause 4.23.4, which carries only NOTE 1.) A LEO round
trip is tens of milliseconds — TS 23.501 NOTE 17 gives roughly 21 ms one way at 1200 km and 13 ms at
600 km — so the base values are conformant there, and extending them would delay every recovery on a
constellation that does not need it. `NR(OTHER_SAT)` is unnamed by that note and takes the base values
by the letter of the specification.

So `UsesExtendedNasSmTimers` is deliberately narrower than `IsNtn`, and a test pins that one is a
strict subset of the other:

```go
if ue.UsesExtendedNasSmTimers() && !ue.IsNtn() { ... }   // must never happen
leo := &AmfUe{RatType: models.RATTYPE_NR_LEO}
if !leo.IsNtn() || leo.UsesExtendedNasSmTimers() { ... } // LEO: NTN, but base timers
```

That relationship is pinned rather than assumed because deriving either predicate from the other is
the mistake worth preventing, and it is an easy one to make in a later cleanup.

## Where the orbit comes from

`ue.RatType`, upgraded at registration from the `RATInformation` the serving RAN advertised for the
TAC at NGSetup. That path already exists in `gmm/handler.go` and this change only reads its result.

## This is unexercised in every deployment I can reach, and that is worth saying plainly

**No RAN available to me advertises `RATInformation`**, so `ue.RatType` stays generic NR, the
indication stays false, and nothing changes. Checked rather than assumed:

- **gnbsim** contains no reference to `RATInformation` anywhere.
- **UERANSIM** builds `SupportedTAItem` with a TAC and a broadcast PLMN list only, and never
  populates `iE_Extensions` anywhere in the gNB. It does carry an orbit internally — a hardcoded
  `ratType = 4` (NR-GEO) in the RRC UE context — and logs `RAT type: NR-GEO (value=4)` for every UE,
  which reads exactly like a RAN advertising its orbit. That value never reaches the wire.

That last detail cost me a wrong conclusion before I read the source, so it is recorded here: a log
line that looks like confirmation was the opposite of it.

The mechanism is correct for a Rel-18 RAN and does nothing until one appears. Where nothing
advertises the IE, a deployment runs on configured timer values, which is what the SMF side already
supports. I would rather ship the conformant path and say it is dormant than derive the orbit from
something cheaper and be wrong for LEO.

## Sent only when true

The field is optional, and an SMF that reads it treats absence and false alike. An explicit false
would add a field to every session-create request on every deployment in order to say nothing.

## Tests

Seven RAT types across the predicate — MEO and GEO true, LEO, OTHER_SAT, NR, EUTRA and unset false —
plus the subset relationship above. Mutation-verified: substituting `IsNtn` for the predicate fails
the LEO case.

`go build ./...`, `go test ./context/... ./consumer/... ./gmm/... -race` and `golangci-lint run` are
clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GyNr6vp6JaxxzPyVcuHXTf

### DIFF
--- a/consumer/sm_context.go
+++ b/consumer/sm_context.go
@@ -301,6 +301,18 @@ func buildCreateSmContextRequest(ue *amf_context.AmfUe, smContext *amf_context.S
 	if ue.RatType != "" {
 		smContextCreateData.SetRatType(ue.RatType)
 	}
+	// TS 24.501 subclause 4.23.4: tell the SMF when this UE's access warrants the extended NAS
+	// session management timer values, so its timers are chosen from the access rather than from
+	// configuration alone. Only NR(MEO) and NR(GEO) qualify — see UsesExtendedNasSmTimers, which
+	// is deliberately narrower than "is this non-terrestrial".
+	//
+	// Sent only when true. The field is optional and an SMF that reads it treats absence and false
+	// alike, so an explicit false would add a field to every request on every deployment to say
+	// nothing.
+	if ue.UsesExtendedNasSmTimers() {
+		smContextCreateData.SetExtendedNasSmTimerInd(true)
+		ue.GmmLog.Infof("signalling the extended NAS SM timer indication to the SMF for RAT type %s", ue.RatType)
+	}
 	// Forward the UE location to the SMF at session creation, mirroring the
 	// SmContextUpdateData paths that already do so. ue.Location is populated
 	// during registration (gmm/handler.go) from NGAP UpdateLocation.
--- a/consumer/sm_context.go
+++ b/consumer/sm_context.go
@@ -301,10 +301,13 @@ func buildCreateSmContextRequest(ue *amf_context.AmfUe, smContext *amf_context.S
 	if ue.RatType != "" {
 		smContextCreateData.SetRatType(ue.RatType)
 	}
-	// TS 24.501 subclause 4.23.4: tell the SMF when this UE's access warrants the extended NAS
-	// session management timer values, so its timers are chosen from the access rather than from
-	// configuration alone. Only NR(MEO) and NR(GEO) qualify — see UsesExtendedNasSmTimers, which
-	// is deliberately narrower than "is this non-terrestrial".
+	// TS 24.501 subclause 4.23.4: "If the use of extended NAS timer for access via a satellite
+	// NG-RAN cell is indicated by the AMF ... the SMF shall calculate the value of the applicable
+	// NAS timer indicated in table 10.3.2 for access via a satellite NG-RAN cell." This is that
+	// indication, so the SMF's timers follow the access rather than configuration alone.
+	//
+	// Which accesses qualify is table 10.3.2 NOTE 5, not the subclause: NR(MEO) and NR(GEO) only.
+	// See UsesExtendedNasSmTimers, which is deliberately narrower than "is this non-terrestrial".
 	//
 	// Sent only when true. The field is optional and an SMF that reads it treats absence and false
 	// alike, so an explicit false would add a field to every request on every deployment to say
--- a/context/amf_ue.go
+++ b/context/amf_ue.go
@@ -572,11 +572,15 @@ func (ue *AmfUe) CmIdle(anType models.AccessType) bool {
 // UsesExtendedNasSmTimers reports whether this UE's access warrants the extended NAS session
 // management timer values, which the AMF signals to the SMF as ExtendedNasSmTimerInd.
 //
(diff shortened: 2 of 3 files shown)


consumer/sm_context.go
```diff
@@ -301,6 +301,18 @@ func buildCreateSmContextRequest(ue *amf_context.AmfUe, smContext *amf_context.S
 	if ue.RatType != "" {
 		smContextCreateData.SetRatType(ue.RatType)
 	}
+	// TS 24.501 subclause 4.23.4: tell the SMF when this UE's access warrants the extended NAS
```
gab-arrobo: @midwell, are you sure the reference is correct? I only see a single `note` (NOTE 1) in that section

```
4.23.4	5GS session management via a satellite NG-RAN cell
For 5GS session management via a satellite NG-RAN cell the UE shall apply the value of the applicable NAS timer indicated in table 10.2.1 for access via a satellite NG-RAN cell.
NOTE 1:	The applied NAS timer values are based on the current satellite NG-RAN access RAT type determined based on information from lower layers.
The NAS timer value obtained is used as described in the appropriate procedure subclause of this specification. The NAS timer value shall be calculated at start of a NAS procedure, and shall not be re-calculated until the NAS procedure is completed, restarted or aborted.
If the use of extended NAS timer for access via a satellite NG-RAN cell is indicated by the AMF (see 3GPP TS 23.501 [8] and 3GPP TS 23.502 [9]), the SMF shall calculate the value of the applicable NAS timer indicated in table 10.3.2 for access via a satellite NG-RAN cell.
The NAS timer value obtained is used as described in the appropriate procedure subclause of this specification. The NAS timer value shall be calculated at start of a NAS procedure and shall not be re-calculated until the NAS procedure is completed, restarted or aborted.
```

gab-arrobo (changes requested):
Please validate that the reference is correct to support the changes. Thanks!

midwell:
You're right, and thank you — the reference was wrong. Subclause 4.23.4 carries only NOTE 1, as you say.

The statement this change relies on — that the satellite timer values apply when the satellite NG-RAN RAT type is NR(MEO) or NR(GEO) — is **NOTE 5 of table 10.3.2, *Timers of 5GS session management – SMF side***. It is the note attached to T3591's own row there: base `16s`, `For access via a satellite NG-RAN cell, 22s`, NOTE 5. Table 10.3.1 says the same for the UE side in its NOTE 7.

Subclause 4.23.4 is still the right citation for the mechanism, so it stays — but quoted rather than paraphrased, since your quote of it is exactly the sentence that matters: *"If the use of extended NAS timer for access via a satellite NG-RAN cell is indicated by the AMF … the SMF shall calculate the value of the applicable NAS timer indicated in table 10.3.2 for access via a satellite NG-RAN cell."* This change is that indication; the table note is what scopes which accesses get it. I had conflated the two.

Corrected in `7375ab6` — three code comments and the PR description. Comments only, no behaviour change, and rebuilt on your rebase rather than over it.

gab-arrobo (approved):

PR #813, opened 2026-09-07
feat(metrics): report the NGAP association count and last-message time

midwell:
## Why

An AMF that has lost every SCTP association keeps running, answers on its SBI, and logs nothing.
From outside the pod it is indistinguishable from an idle one. We spent thirty minutes of a
10 000-UE run in exactly that state — all four associations gone within 2 ms of each other with
`ETIMEDOUT`, `restartCount` still 0, the listener still bound on 38412, `/proc/net/sctp/assocs`
empty — and nothing the AMF publishes could have shown it. The run appeared to be progressing.

## What this adds

Two gauges, both on the existing `:9089` metrics endpoint:

- **`amf_ngap_associations`** — how many SCTP associations this AMF currently terminates, updated
  where the listener registers an accepted association and where the connection handler drops it.
- **`amf_ngap_last_message_timestamp_seconds`** — when an NGAP message was last **received and
  decoded**, stamped on both intake paths (`Dispatch` for direct SCTP and `DispatchLb` for the
  load-balanced one). The wording is deliberate: the stamp sits after the decode and before the
  message is queued, so a stalled *handler* behind a live intake still advances it, and a message
  that fails to decode does not. "Handled" would have overstated it in exactly the fault this is
  meant to expose.

They are useful together rather than separately: the count going to zero means no RAN, while the
timestamp standing still with a non-zero count means a RAN is attached and nothing is being
processed. Those are different faults with the same symptom, and today neither is visible.

## Two things worth stating about the count

- **It comes from the listener's own map, not from `AmfRanPool`.** That pool is a single
  `sync.Map` keyed three ways — by `net.Conn`, by remote address, and by GnbId string — so its
  length counts keys rather than associations, and it is not a count of anything in particular.
- **With the SCTP load balancer in use it stays 0, and the Help string says why.** The listener
  is started either way — `ngap_service.Run` is not conditional on the mode — so this is not a
  matter of the listener being absent: the gNBs connect to the load balancer, so this AMF accepts
  nothing. The gauge counts what this process terminates rather than what it serves, and a
  dashboard should not read it as "no RAN" in that deployment. The last-message timestamp stays
  meaningful there, since the load-balanced dispatcher stamps it too.

The count and its publish are taken under one lock, so two associations changing at once cannot
leave the gauge holding the earlier of the two counts until the next event moves it.

## Deliberately not here

The health signal these numbers exist to support. A liveness assertion — that an AMF which has
held an association and now holds none is unhealthy — needs a latch to distinguish "never served"
from "stopped serving", and that is a behaviour change rather than an observation. It follows in
its own pull request; this one only makes the state visible.

## Verification

All on linux/amd64, in a container at CI's pinned versions:

- `go build ./...`, `go vet ./...` clean; `go test ./... -race -count=1` ok.
- `golangci-lint run --config .golangci.yml ./...` at **v2.13.2**: 0 issues; `golangci-lint fmt
  --diff` empty.
- New tests in `ngap/service/metrics_test.go` drive the **reporting helper** through both
  directions — none, one, two, one, none — using `net.Pipe` associations so the test needs no
  socket, and assert that the timestamp gauge is *registered*, because a gauge that is set but
  never registered reads as absent rather than as wrong, which is the harder failure to notice.
  The assertions gather from the default registry through the gathered types' accessors, so they
  add no module dependency.

  **Being exact about what that does not cover:** the tests call the helper themselves, so
  removing either production call site — the accept path or the `handleConnection` defer — would
  leave them green. `handleConnection` takes a `*sctp.SCTPConn`, so the call sites cannot be
  driven from `net.Pipe`; exercising them needs a real listener and a kernel with SCTP, which is
  what the sibling shutdown change's test does. I considered a `var publish = reportAssociationCount`
  indirection to make the defer assertable and judged it more machinery than this change earns.

## Two properties of the count, for whoever builds on it

Neither is a defect here, and both matter to anything that reads this gauge as a health signal:

- **It tracks the handler goroutine's lifetime, not the kernel's view of the association.** The
  count drops when `SCTPRead` returns a terminal error. A graceful peer shutdown is immediate; a
  peer that goes silent without closing is noticed only when SCTP heartbeats give up, which is
  minutes at kernel defaults. Anything deriving an alarm from this inherits that latency.
- **A listener that never binds leaves the gauge at 0**, which is indistinguishable from "no RAN
  has connected yet". `listenAndServe` logs and returns if `Listen` fails, so an AMF that can
  never serve publishes the same value as one that is merely idle.

### DIFF
--- a/metrics/telemetry.go
+++ b/metrics/telemetry.go
@@ -14,7 +14,6 @@ package metrics
 import (
 	"encoding/hex"
 	"net/http"
-	"time"
 	"unicode/utf8"
 
 	"github.com/omec-project/amf/logger"
@@ -60,9 +59,10 @@ func initAmfStats() *AmfStats {
 
 		ngapLastMessage: prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: "amf_ngap_last_message_timestamp_seconds",
-			Help: "Unix time of the last NGAP message this AMF handled, or zero if it has " +
-				"handled none. Exposed as a timestamp rather than an age so that the staleness " +
-				"is computed at query time.",
+			Help: "Unix time of the last NGAP message this AMF received and decoded, or zero " +
+				"if it has received none. It advances at intake, before the message is " +
+				"handled, so a stalled handler behind a live intake still moves it. Exposed " +
+				"as a timestamp rather than an age so that staleness is computed at query time.",
 		}),
 	}
 }
@@ -143,11 +143,12 @@ func SetNgapAssociations(count int) {
 	amfStats.ngapAssociations.Set(float64(count))
 }
 
-// SetNgapLastMessage records that an NGAP message was handled at t. Paired with the
-// association count, a timestamp that stops advancing while the count is non-zero is a
-// different fault from the count going to zero, and the two are worth telling apart.
-func SetNgapLastMessage(t time.Time) {
-	amfStats.ngapLastMessage.Set(float64(t.Unix()))
+// SetNgapLastMessage records that an NGAP message has just been received and decoded.
+// Paired with the association count, a timestamp that stops advancing while the count is
+// non-zero is a different fault from the count going to zero, and the two are worth
+// telling apart.
+func SetNgapLastMessage() {
+	amfStats.ngapLastMessage.SetToCurrentTime()
 }
 
 // IncrementDbWriteDropped increments the counter of UE context writes dropped
--- a/metrics/telemetry.go
+++ b/metrics/telemetry.go
@@ -14,6 +14,7 @@ package metrics
 import (
 	"encoding/hex"
 	"net/http"
+	"time"
 	"unicode/utf8"
 
 	"github.com/omec-project/amf/logger"
@@ -26,6 +27,8 @@ type AmfStats struct {
 	ngapMsg           *prometheus.CounterVec
 	gnbSessionProfile *prometheus.GaugeVec
 	dbWriteDropped    prometheus.Counter
+	ngapAssociations  prometheus.Gauge
+	ngapLastMessage   prometheus.Gauge
 }
 
 var amfStats *AmfStats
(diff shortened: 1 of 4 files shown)


midwell:
The `e2e-tests` failure here is not from this change, and the evidence is another run of the same job on a branch that cannot affect NGAP at all.

Run [34153462482](https://github.com/omec-project/amf/actions/runs/34153462482) failed with:

```
Profile Name: profile2, Profile Type: pdusessest
Ue's Passed: 4, Ue's Failed: 1
imsi:imsi-208930100007510, procedure:PDU-SESSION-ESTABLISHMENT-PROCEDURE, error:triggering event:PDU-SESSION-ESTABLISHMENT-REQUEST-EVENT, expected event:PDU-SESSION-ESTABLISHMENT-ACCEPT-EVENT, received event:PDU-SESSION-ESTABLISHMENT-REJECT-EVENT
Profile Status: FAIL
```

Run [34081740315](https://github.com/omec-project/amf/actions/runs/34081740315), which is the dependabot bump of the golang base image from 1.27.0 to 1.27.1, failed **byte-identically** earlier today: same profile, same 4-passed/1-failed split, and the same IMSI `imsi-208930100007510` receiving a PDU session establishment reject.

This change adds two gauges and one timestamp write. It alters no NGAP or session behaviour, and the two sibling pull requests opened from the same base today (#811, #812) both passed the same job. Happy to rebase if that helps trigger a fresh run.

midwell:
One more data point, which arrived after the comment above: #814 is stacked on this branch, so its head **contains this commit**, and its `e2e-tests` run passed. Same code plus a health endpoint, same job, green.

Together with the identical failure on the golang base-image bump ([34081740315](https://github.com/omec-project/amf/actions/runs/34081740315) — same profile, same 4/1 split, same IMSI), that puts the failure here on the job rather than on the change.

gab-arrobo:
> The `e2e-tests` failure here is not from this change, and the evidence is another run of the same job on a branch that cannot affect NGAP at all.
> 
> Run [34153462482](https://github.com/omec-project/amf/actions/runs/34153462482) failed with:

Agree, the failure is NOT because of these changes. It an issue that started showing up in some E2E tests/PRs

gab-arrobo (approved):

PR #819, opened 2026-09-11
fix(ngap): replace the supported TA list on a RAN configuration update

midwell:
## The defect

`HandleRanConfigurationUpdate` builds `ran.SupportedTAList` with the same loop as `HandleNGSetupRequest`, minus one step: NG Setup clears the list first, the configuration update does not. So the entries accumulate.

TS 38.413 clause 8.7.2.2 is explicit about which of the two is right:

> If the Supported TA List IE is included in the RAN CONFIGURATION UPDATE message, the AMF shall **overwrite the whole list of supported TAs** and the corresponding list of supported slices for each TA, and use them for subsequent registration area management of the UE.

Two consequences, both observed by removing the fix and running the tests:

- **A tracking area the gNB has stopped broadcasting stays in the AMF's idea of what this RAN serves.** After an update announcing `000001` and then one announcing `000002`, the list holds `[000001 000002]`. That list is what `InTaiList` is asked about for registration area management, so the AMF goes on treating a tracking area as served after the gNB said it is not.
- **Once the list is full, a genuinely new tracking area is dropped in silence.** The append is guarded by the list's capacity, `maxNumOfTAI × maxNumOfBroadcastPLMNs` = 192, and the guard's `else` is a bare `break`. Driving 200 updates that each announce the same single TAC leaves 192 identical entries, after which nothing new can be recorded and nothing says so.

A gNB that never changes its configuration is unaffected, which is why this has gone unnoticed: the list is only rebuilt when the message carries the IE at all.

## Replacing the list is only half of it

`gnb_session_profile` is written by `SetRanStats`, which walks the RAN's **current** list, and its only callers are the connect transition (`ngap/message/send.go`) and RAN removal (`context/amf_ran.go`) — never a configuration update. That makes the correct list replacement, on its own, a regression:

- **before:** the departing tracking area stayed in the list, so the disconnect walk eventually zeroed its series — stale while the gNB was up, corrected when it left
- **after, without this second half:** the tracking area is gone from the list, so nothing ever visits that series again and it holds `Connected=1` for the life of the process

So the series belonging to the tracking areas being dropped are retired as part of the replacement. They are **deleted rather than zeroed**, because this gNB is still connected: neither state the label carries is true of a tracking area it no longer serves. This is the shape named in `SetRanStats`'s own comment — *"the state is a label, so writing only the series for the state being entered leaves the series for the state being left at whatever it last held"* — at TAC granularity.

Two mechanics worth naming for review:

- The retirement runs **after** the RAN's write lock is released. The rebuild happens inside a closure holding `LockRanState()`, and `SetRanStats`/`statsSnapshot` take `RLockRanState()`; a Go `RWMutex` is not reentrant, so retiring from inside that closure would deadlock the NGAP handler goroutine. The departing list is copied under the lock and used after it.
- Newly announced tracking areas are **not** published by this path — they get their series at the next connect transition, as they did before. That is unchanged behaviour, not something this change introduces.
- `HandleNGSetupRequest` clears the list in the same way and likewise does not retire what it drops, so the same residue is possible there — it needs a second NG Setup on one association carrying a different TA list, where a fresh association instead gets its series zeroed on disconnect while the old list is still intact. Left for its own change rather than widened into this one; `RetireDepartedTacs` is written so it can serve both.

## The change

Clear the list before the loop, exactly as `HandleNGSetupRequest` does, with the clause that requires it named in the comment.

## Tests

`ngap/ran_configuration_update_test.go`, driving the handler rather than the list:

- `TestRanConfigurationUpdateReplacesTheSupportedTaList` — two updates announcing different TACs; the second must leave exactly the second
- `TestRepeatedRanConfigurationUpdatesDoNotFillTheTaList` — 200 updates announcing the same TAC must leave one entry, not 192. It is strictly weaker than the first test and is kept for the capacity story rather than for a property the first does not cover
- `TestARetiredTrackingAreaLeavesTheGauge` — after an update drops a tracking area, no `gnb_session_profile` series for it remains under that gNB's labels

Each was checked against a mutation of the code it covers, in a copy of the tree rather than by assertion. Removing the clear:

```
after the second update the TA list is [000001 000002], want exactly [000002] - the list is overwritten, not added to
after 200 updates the TA list holds 192 entries ([000001 000001 ...]), want 1
```

and removing the retirement:

```
gnb_session_profile still carries a series for 000001 after the gNB stopped broadcasting it, so exported state outlives the condition it describes
```

Both updates in the first test name tracking areas the AMF serves, so each takes the acknowledged path — clause 8.7.2.2's "shall overwrite" is in *Successful Operation*, and a test that pinned the replacement against an update the AMF refused would be pinning something the clause does not say.

The gauge readback matches on the gNB's `id` and `ip` labels as well as `tac`: the registry is global to the package and another test there publishes this gauge for its own gNB with the same TAC, which is a `-count=2` failure rather than a `-count=1` one.

One note for whoever writes the next test here: the bare `context.NewAmfRanDefault()` does not give the RAN its `SupportedTAList`, while every production constructor in `context` does — and since the append is guarded by that list's capacity, a test that skips it records nothing at all and passes for the wrong reason. These tests initialise it the way production does.

## Verification

`pre-commit run --all-files` is clean on darwin with `SKIP=go-test-race,golangci-lint-full,staticcheck` (`ngap/service` needs Linux SCTP, so the module does not build here). Those three ran on linux/amd64 in the `golangci/golangci-lint:v2.13.2` image at the versions `.pre-commit-config.yaml` pins: `golangci-lint run` (0 issues), `staticcheck ./...` v0.8.1 (clean), `go test -race ./... -count=1` (13 packages ok) and `-count=2` over `ngap`, `context` and `metrics`, plus `go build ./...` and `go vet ./...`.


midwell:
Good catch, and taken in `3138788` — the mechanism is real. `go DispatchNgapMsg` at `ngap/dispatcher.go:132` and `:199` puts every NGAP message on its own goroutine, so a configuration update does run alongside an NG Setup response or a disconnect, and both paths snapshot the supported TA list under the read lock and then write metrics after releasing it. An update landing between another goroutine's snapshot and its writes lets that goroutine republish exactly the series the update retired.

A per-RAN mutex now spans each read-and-write pair. Lock order was checked before adding it, since the retirement already had to be moved out of the RAN write lock to avoid a deadlock: both `SetRanStats` callers are outside the state lock — `Remove` does not hold it, and `SendNGSetupResponse` is called after `RUnlockRanState` at `ngap/handler.go:731` — so the new lock is always taken before the state lock, never inside it.

**I could not reproduce the interleaving, and I would rather say so than imply the test pins it.** With the mutex removed: 120 tracking areas, 16 concurrent publishers, 20 update cycles, race detector, five runs — no failures. There is a structural reason. Both loops walk the list in the same order, and the publisher must have snapshotted before the update replaced the list, so it is already ahead of the retirement in that order; for it to write a TAC after the retirement deletes it, it has to be preempted for longer than the whole retirement takes. Real, but rare.

So the added test drives both paths concurrently under `-race` and checks the end state, and its comment says explicitly that removing the lock does not make it fail. The fix stands on construction — one uncontended mutex making the pair atomic — rather than on a test I did not manage to write.

Verified on the pushed head in the linux/amd64 `golangci-lint:v2.13.2` image: `go build ./...`, `go vet ./...`, `golangci-lint run` (0 issues), `staticcheck ./...` v0.8.1, `go test -race -count=1 ./...` (13 packages ok) and `-count=2 ./ngap/`.

ngap/handler.go
```diff
@@ -4256,6 +4260,18 @@ func HandleRanConfigurationUpdate(ran *context.AmfRan, message *ngapType.NGAPPDU
...
+		// supported TAs and the slices of each. Without this the entries accumulate, so a
+		// tracking area the gNB has stopped broadcasting stays in the AMF's idea of what
+		// this RAN serves - and once the list reaches its capacity, a genuinely new one is
+		// dropped instead. HandleNGSetupRequest already clears in the same way.
+		if len(ran.SupportedTAList) != 0 {
+			departedTAList = make([]context.SupportedTAI, len(ran.SupportedTAList))
+			copy(departedTAList, ran.SupportedTAList)
+			ran.SupportedTAList = context.NewSupportedTAIList()
+		}
+
```
gab-arrobo, line 4274: I think this update should happen only if the RAN configuration update is successful. The way it is done in this PR, the `SupportedTAList` gets updated regardless the procedure is successful or not. What do you think?

I think the following changes are needed:
```diff
diff --git a/ngap/handler.go b/ngap/handler.go
index 93e4e40..98cda2c 100644
--- a/ngap/handler.go
+++ b/ngap/handler.go
@@ -4260,18 +4260,12 @@ func HandleRanConfigurationUpdate(ran *context.AmfRan, message *ngapType.NGAPPDU
                        return true
                }

-               // The list is replaced, not added to: TS 38.413 clause 8.7.2.2 says that when the
-               // Supported TA List IE is included, the AMF shall overwrite the whole list of
-               // supported TAs and the slices of each. Without this the entries accumulate, so a
-               // tracking area the gNB has stopped broadcasting stays in the AMF's idea of what
-               // this RAN serves - and once the list reaches its capacity, a genuinely new one is
-               // dropped instead. HandleNGSetupRequest already clears in the same way.
-               if len(ran.SupportedTAList) != 0 {
-                       departedTAList = make([]context.SupportedTAI, len(ran.SupportedTAList))
-                       copy(departedTAList, ran.SupportedTAList)
-                       ran.SupportedTAList = context.NewSupportedTAIList()
-               }
-
+               // Built into a candidate first, not written straight into ran.SupportedTAList: TS
+               // 38.413 8.7.2.2 frames "the AMF shall overwrite the whole list of supported TAs"
+               // as part of the AMF having "successfully updated the configuration data", so a
+               // candidate the checks below still reject must leave the previous list, and its
+               // exported metrics, untouched.
+               candidateTAList := context.NewSupportedTAIList()
                for i := 0; i < len(supportedTAList.List); i++ {
                        supportedTAItem := supportedTAList.List[i]
                        tac := hex.EncodeToString(supportedTAItem.TAC.Value)
@@ -4286,7 +4280,7 @@ func HandleRanConfigurationUpdate(ran *context.AmfRan, message *ngapType.NGAPPDU
                                        ratInformation = &ngapType.RATInformation{Value: found.Value}
                                }
                        }
-                       capOfSupportTai := cap(ran.SupportedTAList)
+                       capOfSupportTai := cap(candidateTAList)
                        for j := 0; j < len(supportedTAItem.BroadcastPLMNList.List); j++ {
                                supportedTAI := context.NewSupportedTAI()
                                supportedTAI.Tai.Tac = tac
@@ -4314,15 +4308,15 @@ func HandleRanConfigurationUpdate(ran *context.AmfRan, message *ngapType.NGAPPDU
                                        }
                                }
                                ran.Log.Debugf("PLMN_ID %+v TAC %s", plmnId, tac)
-                               if len(ran.SupportedTAList) < capOfSupportTai {
-                                       ran.SupportedTAList = append(ran.SupportedTAList, supportedTAI)
+                               if len(candidateTAList) < capOfSupportTai {
+                                       candidateTAList = append(candidateTAList, supportedTAI)
                                } else {
                                        break
                                }
                        }
                }

-               if len(ran.SupportedTAList) == 0 {
+               if len(candidateTAList) == 0 {
                        ran.Log.Warnln("RanConfigurationUpdate failure: No supported TA exist in RanConfigurationUpdate")
                        cause.Present = ngapType.CausePresentMisc
                        cause.Misc = &ngapType.CauseMisc{
@@ -4336,7 +4330,7 @@ func HandleRanConfigurationUpdate(ran *context.AmfRan, message *ngapType.NGAPPDU
                                taiList[i].Tac = util.TACConfigToModels(taiList[i].Tac)
                                ran.Log.Infof("Supported Tai List in AMF Plmn: %v, Tac: %v", taiList[i].PlmnId, taiList[i].Tac)
                        }
-                       for i, tai := range ran.SupportedTAList {
+                       for i, tai := range candidateTAList {
                                if context.InTaiList(tai.Tai, taiList) {
                                        found = true
                                        ran.Log.Debugf("SERVED_TAI_INDEX[%d]", i)
@@ -4352,6 +4346,24 @@ func HandleRanConfigurationUpdate(ran *context.AmfRan, message *ngapType.NGAPPDU
                        }
                }

+               // The list is replaced, not added to, and only now that the candidate is accepted:
+               // TS 38.413 8.7.2.2 - "If the Supported TA List IE is included in the RAN
+               // CONFIGURATION UPDATE message, the AMF shall overwrite the whole list of supported
+               // TAs and the corresponding list of supported slices for each TA, and use them for
+               // subsequent registration area management of the UE" (InTaiList is what consults
+               // this list for that purpose). Without this the entries would accumulate, so a
+               // tracking area the gNB has stopped broadcasting would stay in the AMF's idea of
+               // what this RAN serves - and once the list reached its capacity, a genuinely new one
+               // would be dropped instead. HandleNGSetupRequest already clears in the same way, on
+               // a fresh connection where there is nothing yet to preserve on rejection.
+               if cause.Present == ngapType.CausePresentNothing {
+                       if len(ran.SupportedTAList) != 0 {
+                               departedTAList = make([]context.SupportedTAI, len(ran.SupportedTAList))
+                               copy(departedTAList, ran.SupportedTAList)
+                       }
+                       ran.SupportedTAList = candidateTAList
+               }
+
                sendAcknowledge = cause.Present == ngapType.CausePresentNothing
                return true
        }()
@@ -4361,7 +4373,9 @@ func HandleRanConfigurationUpdate(ran *context.AmfRan, message *ngapType.NGAPPDU

        // Replacing the list is only half of it: the gauge is written from the list, so a
        // tracking area that has just left it would never be written again and would keep its
-       // last sample for ever. Retiring them here, outside the RAN's write lock.
+       // last sample for ever. Retiring them here, outside the RAN's write lock. departedTAList
+       // is only ever populated on the acknowledged path above, so a rejected update is a no-op
+       // here too.
        ran.RetireDepartedTacs(departedTAList)

        if sendAcknowledge {
```

gab-arrobo (commented):
Let me know what you think about this suggestion. Also, please add a test that updates the `candiadateTAList` but takes an unsuccessful path such that we are covered from any future changes that may break this behavior. Thanks!

gab-arrobo:
@midwell, please resolve conflict and rebase PR. Thanks!
image

midwell:
Everything this branch carries is already on `main`. It went in with #821, which was stacked on this branch, so that squash (`c07f378`) contains both of these commits in full — which is why this now shows a conflict rather than having anything left to rebase. Verified by content on `main`: `RetireDepartedTacs` at `context/amf_ran.go:315`, the `statsMu` pairing of "read the list, then write the gauge from it" at `:55-59`, and the replacement plus the retirement in `HandleRanConfigurationUpdate`.

**On the suggestion: I agree with it, and it is on `main` too, as #824.** I had checked it against this branch before that landed — applied here it vets clean and keeps every test in `ran_configuration_update_test.go` green, and the rejected-path test you asked for fails without it on *both* halves: the list ends up as `[000002]`, the tracking area the AMF just refused, and the retirement then deletes the `gnb_session_profile` series for `000001`, which the gNB still broadcasts. The second half is the one that matters most and the one I would have shipped — Copilot named it on line 4273, and your diff is what fixes it.

The spec agrees with that shape. Clause 8.7.2.2 puts "the AMF shall overwrite the whole list of supported TAs" inside *Successful Operation*, next to "the AMF responds with a RAN CONFIGURATION UPDATE ACKNOWLEDGE message to acknowledge that it successfully updated the configuration data", and 8.7.2.3 says only that a refusal is answered with RAN CONFIGURATION UPDATE FAILURE — so a refused update has to leave the stored data alone. The `serveTACs` helper in this branch's test file already carried that reasoning as the reason its updates take the acknowledged path; the candidate list is what makes it true of the code as well.

One note on the asymmetry your diff's comment mentions, since `main` now keeps it: NG Setup differs in the spec rather than by accident. Clause 8.7.1.1 has the procedure "erases any existing application level configuration data in the two nodes, replaces it by the one received" — stated as the procedure's own effect, not as something conditional on success. So it is not quite that there is nothing to preserve on a fresh connection; a second NG Setup on one association does have a previous list, which is what #821 was about. It is that the erase is unconditional there, which is why retiring the departed series on a refused NG Setup is still right.

The rename follow-up named in #821's body is on `main` as well, as #823, with `RetireDepartedTacs` taking the previous name and IP — the signature it needed.

So there is nothing left for this pull request to carry, and I am closing it. Nothing is lost: #821's squash preserved both commit messages from here.

PR #820, opened 2026-09-11
Test shared E2E workflow revision

andybavier:
## Purpose

Validate the current single-image AMF E2E path against the proposed shared E2E workflow implementation in omec-project/.github#204.

This changes only the `e2e-test.yml` reference to commit `799176c831bc7130b5bd7dd5be79d6af559e0259`; all AMF inputs and workflow behavior remain unchanged.

A successful `e2e-tests` run is the backward-compatibility acceptance test for the shared workflow. This temporary PR should not be merged.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,6 +119,6 @@ jobs:
       checks: write
       id-token: write
       attestations: write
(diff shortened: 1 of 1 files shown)


gab-arrobo:
@andybavier, should we close this PR? Thanks!

andybavier:
Shared workflow has merged, not needed any more.

